PR #6, opened 2024-08-19
Fixes fetchAllFilterOptions query structure

Joels-code:
Fixes for the fetchAllFilterOptions method:
1) 400'ing due to an extra '}' in the query -> removed that
2) returning undefined due to `isList` being `false` in the `fetchSanity` method. Setting this to `true` made it return results as expected.
3) The query builder did not account for some possibly undefined params, leading to incorrect fetch results. Adjusted the query builder to return empty strings where params are undefined.

### DIFF
--- a/src/services/sanity.js
+++ b/src/services/sanity.js
@@ -615,15 +615,15 @@ export async function fetchAllFilterOptions(
     const query = `
         {  
           "meta": {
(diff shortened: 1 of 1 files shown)


Joels-code:
@miguelfondeur I request you as a reviewer because AFAIK the FE and MA teams will both be using these methods. Please let me know if these changes are fine for your end as well.

PR #12, opened 2024-08-23
Added method groq

csillaj:


Joels-code (approved):

adrian-musora (approved):

src/contentTypeConfig.js
```diff
@@ -4,6 +4,8 @@ let contentTypeConfig = {
             '"artist_name": artist->name',
             'soundslice',
             'instrumentless',
+            '"id": railcontent_id',
+            '"type": _type',
```
adrian-musora, line 8: These fields are part of the `defaultFields` object, why are they needed here?

csillaj: defaultFields is defined only in fetchAll method, not globally

PR #19, opened 2024-08-27
Adds support for other filter types and adds filtersToGroq function

Joels-code:


src/contentTypeConfig.js
```diff
@@ -238,12 +238,34 @@ function getFieldsForContentType(contentType, asQueryString=true) {
...
+            let [key, value] = field.split(',');
+            switch (key) {
+              case 'difficulty':
+                return `&& difficulty_string == "${value}"`;
+              case 'genre':
+                return `&& genre[]->name match "${value}"`;
+              case 'topic':
+                return `&& topic[]->name match "${value}"`;
+              default:
+                return `&& ${key} == "${value}"`;
```
csillaj, line 257: you could add here case 'instrumentless' because that too is a special case `return && ${key} == ${value}`;` because value is boolean and not string

miguelfondeur (approved):

PR #21, opened 2024-08-30
MA - Added groq for challenge overview

MeenuDhanoa:


src/index.js
```diff
@@ -29,7 +29,7 @@ import {
   fetchChildren,
   fetchParentByRailContentId,
   fetchMethodPreviousNextLesson,
-  fetchLiveEvent,
+  fetchChallengeOverview,
```
adrian-musora, line 32: We need this.
```suggestion
  fetchLiveEvent,
  fetchChallengeOverview,
```

src/index.js
```diff
@@ -72,5 +72,5 @@ export {
     fetchChildren,
     fetchParentByRailContentId,
     fetchMethodPreviousNextLesson,
-    fetchLiveEvent,
+    fetchChallengeOverview,
```
adrian-musora, line 75:
```suggestion
    fetchLiveEvent,
    fetchChallengeOverview,
```

src/services/sanity.js
```diff
@@ -886,6 +886,39 @@ export async function fetchPackChildren(railcontentId) {
   return fetchChildren(railcontentId, 'pack');
 }
 
+export async function fetchChallengeOverview(id) {
```
adrian-musora, line 889: This is an ignorant question that i couldn't find a quick answer to. Is this method intended to only get `Challenge` type content? If so, I'd prefer if we moved the fields to the `contentTypeConfig.js` to maintain the design.

adrian-musora (changes requested):

csillaj (approved):

Joels-code:
I agree with Adrians points. I will re-review once the changes are made

src/contentTypeConfig.js
```diff
@@ -52,6 +52,12 @@ let contentTypeConfig = {
             '"primary_cta_text": select(dateTime(published_on) > dateTime(now()) && dateTime(enrollment_start_time) > dateTime(now()) => "Notify Me", "Start Challenge")',
             'challenge_state',
             'challenge_state_text',
+            'description',
```
adrian-musora, line 55: Description is an object field, so has a complex structure. We may update the format later, but for the moment are using the the raw html. 
```suggestion
            `"description": ${descriptionField}`,
```

adrian-musora (approved):

PR #23, opened 2024-09-03
MA: added necessary fields for routines

csillaj:


src/contentTypeConfig.js
```diff
@@ -227,6 +227,13 @@ let contentTypeConfig = {
            `"description": ${descriptionField}`,
            '"instructors": instructor[]->name'
         ],
+    },
+    'routine':{
+        'fields': [
+            'description',
```
adrian-musora, line 233: Description is a block field and has a strange structure. We may change the Sanity behaviour later, so using the constant should make it more future proof.
```suggestion
            `"description": ${descriptionField}`,
```

adrian-musora (approved):

PR #25, opened 2024-09-04
MA: added groq for foundation

csillaj:


src/contentTypeConfig.js
```diff
@@ -255,7 +255,31 @@ let contentTypeConfig = {
...
-    }
+    },
+    'foundation': {
+        'fields': [
+            `"description": ${descriptionField}`,
+            `"instructors":instructor[]->name`,
+            `published_on`,
+            `"id": railcontent_id`,
+            `title`,
+            `"type": _type`,
```
adrian-musora, line 266: These fields are included in the DEFAULT_FIELDS already. 
It looks like sanity handles merges the duplicated fields, but 
```suggestion
```

adrian-musora (approved):

PR #30, opened 2024-09-06
MA: added groq for packs

csillaj:


src/services/sanity.js
```diff
@@ -831,6 +831,22 @@ export async function fetchRelatedLessons(railContentId, brand) {
...
+* Fetch all packs.
+* @param {string} brand - The brand for which to fetch packs.
+* @param {string} [searchTerm=""] - The search term to filter packs.
+* @param {string} [sort="-published_on"] - The field to sort the packs by.
+* @returns {Promise<Array<Object>|null>} - The fetched pack content data or null if not found.
+*/
+export async function fetchAllPacks(brand, sort = "-published_on", searchTerm = "") {
+  const sortOrder = getSortOrder(sort);
+
+  const query = `*[_type == 'pack' && brand == '${brand}' && title match "${searchTerm}*"]{
```
adrian-musora, line 844: Just FYI. 
I was curious if adding the search `title match "*"` would have any performance effect and it actually improved my test queries. I'm not sure what's going on here, but happy that sanity.

src/services/sanity.js
```diff
@@ -839,18 +855,9 @@ export async function fetchRelatedLessons(railContentId, brand) {
...
-        "artist_name": artist->name,
-        artist,
-        difficulty,
-        difficulty_string,
-        web_url_path,
-        published_on
-      } | order(published_on asc)[0...5]`
-  return fetchSanity(query, true);
+    ${getFieldsForContentType('pack')}
+  } | order(published_on asc)`
```
adrian-musora, line 859: Slicing the array at the request level is a small performance improvement.  
```suggestion
  } | order(published_on asc) [0...1]`
```

adrian-musora (approved):

PR #31, opened 2024-09-06
Fetchall updates

peterminhlee:


adrian-musora (approved):
I can't comment on the line in github, But this will reduce the changes to a single line on line 420
`let fields = useDefaultFields ?  customFields.concat(DEFAULT_FIELDS.concat(additionalFields)) : customFields;`

PR #39, opened 2024-09-12
MA Changes for Coaches

DidierF:


src/contentTypeConfig.js
```diff
@@ -174,9 +174,20 @@ let contentTypeConfig = {
...
+            '"coach_top_banner_image": coach_top_banner_image.asset->url',
+            '"coach_bottom_banner_image": coach_bottom_banner_image.asset->url',
             '"coach_card_image": coach_card_image.asset->url',
-            'focus'
+            '"coach_featured_image": coach_featured_image.asset->url',
+            '"coach_top_banner_image": coach_top_banner_image.asset->url',
+            'focus',
+            'focus_text',
+            'forum_thread_id',
+            '"id": railcontent_id',
```
adrian-musora, line 185: this is already include in the default fields
```suggestion
```

src/services/sanity.js
```diff
@@ -489,6 +489,9 @@ export function getSortOrder(sort= '-published_on', groupBy)
         case "slug":
             sortOrder = groupBy ? 'name' : "title";
             break;
+        case "name":
+            sortOrder = "name";
+            break;
         case "popularity":
             sortOrder = "popularity";
             break;
```
adrian-musora, line 497: This is a smidge cleaner.
```suggestion
        case "name":
        case "popularity":
            sortOrder = sort;
            break;
```

adrian-musora (approved):

PR #42, opened 2024-09-12
fetchLessonContent updates

peterminhlee:


miguelfondeur (approved):

src/services/sanity.js
```diff
@@ -822,7 +822,12 @@ export async function fetchLessonContent(railContentId) {
             "coach_profile_image":thumbnail_url.asset->url
           },
           "instructors":instructor[]->name,
-          instructor[]->,
+          "instructor": instructor[]->{
+            "id":_id,
```
csillaj, line 826: I'll readd short_bio here, because the MA needs that field as well

csillaj: Isn't using railcontent_id for "id" better, instead of _id?

PR #43, opened 2024-09-13
Add Filter Restrictions handler

adrian-musora:
Design:
- queries that require modifications on the filter that were previously handled by ContentService/ContentReposity now can use a `FilterBuilder` instance to handle the creation of the filter arguments. 
- Added ulitity function to build queries raw, or using the filter builder option
- clean up existing methods to use utility functions or explicitly call the generic version of the method
- add (easy) missing tests to check that methods generate valid groq. The tests don't fully test the functionality of the methods, only compilation of groq.

Further Work to do:
- `fetchAll` was not touched. This will require it's own PR to complete because of the complexity. I wanted to get the framework in place to avoid too many merge conflicts/updates. 
- see table below. Some methods need to be refactored or updated

Testing information (what I did)

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head>	
</head>

<body>

Method | Uses New Filter | User Raw Query | Hardcoded | Uses FetchAll | Has Test |  
-- | -- | -- | -- | -- | -- | --
fetchSongById | 1 |   |   |   | 1 |  
fetchArtists | 1 |   |   |   | 1 | This maybe shouldn’t. Not sure what this is used for
fetchSongArtistCount | 1 |   | 1 |   | 1 |  
fetchRelatedSongs |   |   | 1 |   | X | This needs to be refactored
fetchAllSongs |   |   |   | 1 | 1 |  
fetchSongFilterOptions |   | 1 |   |   | 1 | Should be moved to filter I think
fetchSongCount |   |   | 1 |   | 1 |  
fetchWorkouts | 1 |   |   |   | 1 |  
fetchNewReleases | 1 |   |   |   | 1 |  
fetchUpcomingEvents | 1 |   |   |   | 1 |  
fetchByRailContentId |   | 1 |   |   | 1 |  
fetchByRailContentIds |   |   | 1 |   | 1 |  
fetchAll |   |   |   | 1 | 1 | several tests
fetchAllFilterOptions |   |   | 1 |   | 1 | Should be moved to filter I think
fetchFoundation | 1 |   |   |   | 1 |  
fetchMethods |   |   | 1 |   | 1 |  
fetchMethod |   |   | 1 |   | 1 | Should be moved to filter I think
fetchMethodChildren |   |   |   |   | X |  
fetchMethodNextLesson |   |   |   |   | X |  
fetchMethodChildrenIds |   |   |   |   | X |  
fetchNextPreviousLesson |   |   |   |   | X |  
fetchRelatedLessons |   |   | 1 |   | 1 | Should be moved to filter I think
fetchAllPacks | 1 |   |   |   | 1 |  
fetchPackAll | 1 |   |   |   | 1 |  
fetchPackChildren |   |   | 1 |   | X |  
fetchLessonContent | 1 |   |   |   | 1 | Maybe shouldn’t be?
fetchCourseOverview | 1 |   |   |   | 1 |  
fetchChildren |   |   | 1 |   | 1 |  
fetchParentByRailContentId |   |   | 1 |   | 1 |  
fetchMethodPreviousNextLesson |   |   |   |   | X |  
fetchChallengeOverview |   |   | 1 |   | 1 |  
fetchCoachLessons |   | 1 |   |   | 1 |  
fetchByReference |   | 1 |   |   | 1 |  

</body>

</html>


src/filterBuilder.js
```diff
@@ -0,0 +1,135 @@
...
+
+    _orWhere(query) {
+        if (!this.filter) throw new Error("invalid query, _orWhere needs to be called on an existing query");
+        this.filter += ` || (${query})`;
+    }
+
+    _trimAmpersands() {
+        this.filter = this.filter.trim();
+        while( this.filter.charAt(0) === '&' ) this.filter = this.filter.substring(1);
+        while( this.filter.charAt(this.filter.length) === '&' ) this.filter = this.filter.slice(-1);
```
adrian-musora, line 122:
```suggestion
        while( this.filter.charAt(0) === '&'  || this.filter.charAt(0) === ' ' ) this.filter = this.filter.substring(1);
        while( this.filter.charAt(this.filter.length) === '&' || this.filter.charAt(this.filter.length) === ' ' ) this.filter = this.filter.slice(-1);
```

src/services/sanity.js
```diff
@@ -254,23 +260,32 @@ export async function fetchWorkouts(brand) {
...
+      "type": _type,
+      web_url_path,`;
+  const filterParams = {};
+  const query = buildQuery(
+      filter,
+      filterParams,
+      fields,
+      {
+          sortOrder: 'published_on desc',
+          end:5,
```
BrandonKerr, line 287: Shouldn't these use the variables?
```suggestion
          sortOrder: sortOrder,
          end: end,
```

src/services/sanity.js
```diff
@@ -354,6 +377,7 @@ export async function fetchScheduledReleases(brand, { page = 1, limit = 10 }) {
 * Fetch content by a specific Railcontent ID.
 *
 * @param {string} id - The Railcontent ID of the content to fetch.
+ * @param {string} contentType - The document type of content to fetch
```
BrandonKerr, line 380: Really nit picky, but there's an unnecessary space
```suggestion
* @param {string} contentType - The document type of content to fetch
```

test/sanityQueryService.test.js
```diff
@@ -294,3 +333,148 @@ describe('Sanity Queries', function () {
...
+            getFutureScheduledContentsOnly: true});
+        const finalFilter = builder.buildFilter();
+        const clauses = spliceFilterForAnds(finalFilter);
+        expect(clauses[0].phrase).toBe(filter);
+        expect(clauses[1].field).toBe('(status'); // extra ( because it's a multi part filter
+        expect(clauses[1].operator).toBe('in');
+        // this needs to match on the
+        const expected = "['published','unlisted'] || (status == 'scheduled' && published_on >=";
+        console.log(clauses[1].condition);
+        console.log(expected)
```
BrandonKerr, line 380: This needs to match on the ....?
also, commenting out the console logs so it doesn't make extra noise
```suggestion
        // this needs to match on the
        const expected = "['published','unlisted'] || (status == 'scheduled' && published_on >=";
        // console.log(clauses[1].condition);
        // console.log(expected)
```

test/sanityQueryService.test.js
```diff
@@ -294,3 +333,148 @@ describe('Sanity Queries', function () {
...
+                    user: {},
+                    permissions: [91, 92],
+                }});
+        const finalFilter = builder.buildFilter();
+        const expected = "references(*[_type == 'permission' && railcontent_id in [91,92]]._id)"
+        const isMatch = finalFilter.includes(expected);
+        expect(isMatch).toBeTruthy();
+    });
+
+    test('withUserPermissions', async () => {
```
BrandonKerr, line 398: I feel like this should have a unique test name. Maybe something like `withUserPermissionsForPlusUser`?

src/filterBuilder.js
```diff
@@ -0,0 +1,135 @@
...
+        this.availableContentStatuses = availableContentStatuses;
+        this.bypassPermissions = bypassPermissions;
+        this.pullFutureContent = pullFutureContent;
+        this.getFutureContentOnly = getFutureContentOnly;
+        // TODO should getFollowedContentOnly be re-added. Currently prod this is only based on subscribed coaches.
+        // If so update _applyFollowedContentOnly()
+        // this.getFollowedContentOnly = getFollowedContentOnly;
+        this.getFutureScheduledContentsOnly = getFutureScheduledContentsOnly;
+        this.filter = filter;
+        this.debug = true;
```
BrandonKerr, line 28: This would be nice to have read from the .env
```suggestion
        this.debug = process.env.DEBUG === 'true' || false;
```
(note I didn't test this suggested change)

BrandonKerr (approved):
I think this looks great! I left a few small notes, and would want @roxanariza's final approval for the nuances of the ContentRepository logic before merging

src/services/sanity.js
```diff
@@ -64,6 +69,7 @@ export async function fetchArtists(brand) {
 * @returns {Promise<int|null>} - The fetched count of artists.
 */
 export async function fetchSongArtistCount(brand) {
+  const filter = new FilterBuilder()
```
roxanariza, line 72: I don't know if this line of code is useful here.

roxanariza (approved):
The code looks good overall. I compared the logic with the RailContent implementation and noticed a few minor differences in the _applyContentStatuses method: to view future scheduled content, both getFutureScheduledContentsOnly must be set to true, and the 'scheduled' status needs to be included in the statuses. In RailContent, simply setting getFutureScheduledContentsOnly to true is enough. Your modification makes sense to me, and I agree with it.

src/services/sanity.js
BrandonKerr: I wondered about that too 😂 but noticed this in Adrian's table in the PR comment:

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN">

<html>
<head>	
</head>

<body>

Method | Uses New Filter | User Raw Query | Hardcoded | Uses FetchAll | Has Test |  
-- | -- | -- | -- | -- | -- | --
fetchSongArtistCount | 1 |   | 1 |   | 1 |  


</body>

</html>
So I think it still needs to be updated to actually use the filter and to remove the hardcoded line below it

roxanariza: You are right. But it was strange to see that line of code there, especially since I was only focusing on the 'Uses New Filter - true' column in the table :)

adrian-musora: I had originally planned to migrate it, but I think this is used in generic pages, not in catalogues, so it doesn't make sense to filter the extra information. 
It's been removed.

test/sanityQueryService.test.js
adrian-musora: Updated comment.

PR #44, opened 2024-09-16
Dev/rwillems20240910 user context

robwillems:


src/services/railcontent.js
```diff
@@ -261,4 +261,54 @@ function fetchAbsolute(url, params) {
...
+}
+
+export async function fetchUserContext() {
+    let url = `/content/user_data_all`;
+    const headers = {
+        'Content-Type': 'application/json',
+        'X-CSRF-TOKEN': globalConfig.railcontentConfig.token
+    };
+    try {
+        const response = await fetch(url, {headers});
```
csillaj, line 273: fetching urls like this `/content/user_data_all` doesn't work for MA, we need the baseUrl added (that's why we added the fetchAbsolute method)

robwillems: Got it.  I'll make sure those are working for you once I finish it.  I wanted to test on the FE side before passing it over to you guys.  I'll let you know once its ready to integrate.

PR #48, opened 2024-09-18
Web_url_path on group by groq

roxanariza:
We should have different urls for collection pages:
- artists: https://www.musora.com/drumeo/artists/Nirvana?included_fields[]=type,song
- genre: https://www.musora.com/drumeo/genres/Electronic/quick-tips
- instructor: https://www.musora.com/drumeo/coaches/todd-sucherman/31973?included_fields[]=type,Course
Added web_url_path for collection pages
Added slug in contentTypeConfig, for the contents that should have different slug on collection pages

### DIFF
--- a/test/sanityQueryService.test.js
+++ b/test/sanityQueryService.test.js
@@ -296,19 +296,19 @@ describe('Sanity Queries', function () {
 
     test('fetchAll-GroupBy-Genre', async () => {
         let response = await fetchAll('drumeo', 'solo',{groupBy: 'genre'});
-        console.log(response);
+        // console.log(response);
         expect(response.entity[0].web_url_path).toContain('/drumeo/genres/');
     });
 
     test('fetchAll-GroupBy-Artists', async () => {
         let response = await fetchAll('drumeo', 'song',{groupBy: 'artist'});
-        console.log(response);
+        // console.log(response);
         expect(response.entity[0].web_url_path).toContain('/drumeo/artists/');
     });
 
     test('fetchAll-GroupBy-Instructors', async () => {
         let response = await fetchAll('drumeo', 'course',{groupBy: 'instructor'});
-        console.log(response);
+        // console.log(response);
         expect(response.entity[0].web_url_path).toContain('/drumeo/coaches/');
     });
 });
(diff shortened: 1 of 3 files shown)


test/sanityQueryService.test.js
```diff
@@ -293,4 +293,22 @@ describe('Sanity Queries', function () {
         const response = await fetchScheduledReleases('drumeo', {});
         expect(response.length).toBeGreaterThan(0);
     });
+
+    test('fetchAll-GroupBy-Genre', async () => {
+        let response = await fetchAll('drumeo', 'solo',{groupBy: 'genre'});
+        console.log(response);
+        expect(response.entity[0].web_url_path).toContain('/drumeo/genres/');
+    });
```
BrandonKerr, line 301: More of a question to @adrian-musora since he has more experience with the tests here:
1. Is there a format that the test names should follow? e.g. in the Laravel tests, we try to do something `test_order_updated_dispatches_shopify_sync_customer_job_for_post_launch` so that when reading the test output it's clear what the test is doing: "order updated dispatches shopify sync customer job for post launch".
Seeing the other tests above these, I see others like "fetchAll-IncludedFields" so I'll assume these names are fine.
2. Would there be value in checking each entry in the response.entity's web_url_path, or is it sufficient to just check the first entry?

test/sanityQueryService.test.js
```diff
@@ -293,4 +293,22 @@ describe('Sanity Queries', function () {
...
+
+    test('fetchAll-GroupBy-Artists', async () => {
+        let response = await fetchAll('drumeo', 'song',{groupBy: 'artist'});
+        console.log(response);
```
BrandonKerr, line 305: Also nitpicking and I'm not sure if I'm the only one with this opinion, but I would recommend commenting out the `console.log` calls since it makes some noise that can make it harder to track test output.

BrandonKerr (approved):
LGTM, but would love to hear some feedback from my comments

test/sanityQueryService.test.js
adrian-musora: No specific naming convention. I've been using `methodName-paramaterization` as the naming scheme. So this would be a valid name. 
Most of the tests only verify execution and don't really verify the results package contains valid data.

test/sanityQueryService.test.js
roxanariza: I commented out the console.log calls

PR #49, opened 2024-09-18
Add a function for console logging in tests, to keep things cleaner

BrandonKerr:
I think it's nicer to not have console logging when we run our tests, but wanted to make a quick and easy way to enable it when it's desired.


adrian-musora (approved):

roxanariza (approved):
I love the idea

PR #57, opened 2024-09-20
TP-175 - Needs Access logic

adrian-musora:
[MWP PR](https://github.com/railroadmedia/musora-web-platform/pull/1447)

**Design:**
- add decorator equivalent to `fetchSanity` to allow caller to customize the return package.
- add `processNeedAccess` to add the 'needs_access' property to the results
- needs_access should be set to true if the user doesn't have permissions that match the permission of the content.

**Future Improvements for Performance**
- Flip the default value for `processNeedAccess` to false, and only set it to true if the UserIsBasic OR we are accessing content that needs to be locked (eg: playlists).
- potentially pass in the user permissions, instead of grabbing them from the BE. This requires a bigger refactor of how user gets passed or stored for different methods. I'd like to avoid passing them in explicitly on each call.
- Get permissions from the user store once it's available. Waiting on Rob for this framework, then needs to be extended to include permissions.

This could also be handled in the groq query if we dynamically add the field:
`"need_access" : "<logic here to check user permissions>"` but this would need to be tested for performance vs the solution provided here. It can't be statically defined as the user's permissions are calculated at runtime so adding it to contentTypeConfig.js is not possible with the current design.
 
**Testing**
- Tests pass (I had to add a dummy flag for processNeedsAccess, I'm not sure this is a good idea, but adding a test framework for railcontent is a bit overkill)
-- when the user store is in place, adding a test harness for it would be a relatively easy extension in the testsuite.
- Update DevEndpoint.vue with `fetchByRailContentIds([380094,402204]).then(a => console.log('check needAccess list', a)); fetchByRailContentId(402204).then(a => console.log('check needAccess single', a));`
- Update sanity.js -> getUserPermissions `return [91];`
- Verify calling /devendpoint, check the console for the object and verify `needs_access` is set to true for the items that don't have 91 in the `permission_id` property. 
- Log out and call /devendpoint again, all results should have `needs_access` set to true.


alesevero (approved):
Code looks good, but definitely want @BrandonKerr input here

src/services/sanity.js
```diff
@@ -1301,6 +1310,42 @@ export async function fetchSanity(query, isList) {
   }
 }
 
+async function needsAccessDecorator(results)
+{
+    if (globalConfig.sanityConfig.useDummyRailContentMethods) return results;
```
BrandonKerr, line 1315: _Unresolved variable useDummyRailContentMethods_ - I assume this just needs to be added to the doc block in config.js?

src/services/sanity.js
```diff
@@ -1301,6 +1310,42 @@ export async function fetchSanity(query, isList) {
...
+        results.forEach((result) => {
+            result['need_access'] = doesUserNeedAccessToContent(result, userPermissions);
+        });
+    } else {
+        results['need_access'] = doesUserNeedAccessToContent(results, userPermissions);
+    }
+    return results;
+}
+
+function doesUserNeedAccessToContent(result, userPermissions)
```
BrandonKerr, line 1328: Nitpicking, but is there any sort of structure we could use the the parameters, so that we know `result.permission_id` is valid, etc.?

src/services/sanity.js
```diff
@@ -1301,6 +1310,42 @@ export async function fetchSanity(query, isList) {
...
+            return false;
+        }
+    }
+    return true;
+}
+
+async function getUserPermissions()
+{
+    const permissions = await fetchUserPermissions();
+    return permissions;
```
BrandonKerr, line 1345: Some more nitpicking, but to be a bit cleaner:
```suggestion
    return await fetchUserPermissions();
```

src/services/sanity.js
```diff
@@ -1249,21 +1250,27 @@ export async function fetchGenreLessons(brand, name, contentType, {
...
  * fetchSanity(query, true)
  *   .then(data => console.log(data))
  *   .catch(error => console.error(error));
  */
-export async function fetchSanity(query, isList) {
+
+export async function fetchSanity(query,
+                                  isList,
+                                  { customPostProcess = null,
+                                    processNeedAccess = true,} = {}
```
BrandonKerr, line 1271: I'm not a big fan of this style, to be honest, but I really like the way you can leverage it to use named optional params, so well done there. I think it just highlights how TypeScript is superior to JavaScript 😄

test/sanityQueryService.test.js
```diff
@@ -72,7 +76,27 @@ describe('Sanity Queries', function () {
...
+          }`
+        const newSlug = 'keysmash1';
+        const newField = 1
+        const postProcess = (result) => {
+            result['new_field'] = newField;
+            result['slug'] = newSlug;
+            return result;
+        };
+        const response = await fetchSanity(query, false, {customPostProcess: postProcess});
+        //console.log(response);
```
BrandonKerr, line 92: Just to be consistent
```suggestion
        log(response);
```

BrandonKerr (changes requested):
This looks good, I just have a few comments with varying degrees of nitpicking

src/services/sanity.js
adrian-musora: Yea, I'm in the same boat. It's a workaround for a way that other languages just handle nicely.

src/services/sanity.js
adrian-musora: Tested for custom field or if premission_id is empty/null requests `needs_access` will always be true

BrandonKerr (approved):

PR #61, opened 2024-09-25
TCH-22 Add new tabs for challenges

adrian-musora:
**Design:**
- moved contentMetaData into a separate file
- refactored to have common elements in the `commonMetaData` object to avoid duplication
- added Challenges Tabs
-- these don't work as they require `fetchAll` to be refactor to handle group by on non-references.

**Testing**
- added MetaDataTests, didn't add anything for challenges.


src/contentMetaData.js
```diff
@@ -0,0 +1,1154 @@
...
+            allowableFilters: ['genre', 'focus'],
+            sortBy: '-published_on',
+    },
+    'challenge': {
+        name: 'Challenges',
+            icon: 'icon-courses',
+            description: "... ",
+            allowableFilters: ['difficulty', 'topic', 'genre'],
+            sortBy: '-published_on',
+            modalText: 'Challenges are a collection of Workout-style videos that build your skills one step at a time. They help you develop broader musical skills at a manageable pace — usually over a few days.',
```
MeenuDhanoa, line 18:
```suggestion
            modalText: 'Challenges are a series of guided lessons designed to build your skills day-by-day.',
```

MeenuDhanoa (approved):

src/contentMetaData.js
```diff
@@ -0,0 +1,1154 @@
...
+                },
+                {
+                    name: 'Genres',
+                    short_name: 'Genres',
+                    is_group_by: true,
+                    value: 'genre',
+                },
+            ],
+        },
+        'challenges': {
```
MeenuDhanoa, line 298: I believe this isn’t needed anymore since we’ve added the 'challenge' data type. Both would serve the same purpose.

adrian-musora: These are a distinct content type specifically for drumeo. Roxana would be a better person to ask regarding history. For the moment these will stay in the system, though I may remove them when we do the data import/update later depending on how things fit together.

src/contentMetaData.js
```diff
@@ -0,0 +1,1154 @@
...
+                    value: 'genre',
+                },
+            ],
+        },
+    },
+    'guitareo': {
+        'instructor': {
+            description: "Tackle your next guitar goal with bite-sized courses from many of the world's best guitarists.",
+        },
+        'courses': {
```
roxanariza, line 882: Invalid key. Should remain ''course' like on drumeo/pianote

src/contentMetaData.js
```diff
@@ -0,0 +1,1154 @@
...
+        'student-review': {
+            thumbnailUrl: 'https://d1923uyy6spedc.cloudfront.net/student-reviews-singeo.png',
+            description: "Want feedback on your playing? Submit a video for student review. We will watch your submission and then provide helpful encouragement and feedback. This is a great way to build accountability and benefit from the expertise of our teachers.",
+        },
+    },
+    'singeo':{
+        'instructor': {
+            description: "Your singing journey is unique. You need personalized coaching that helps you reach your goals. Learn from some of the best singers and vocal coaches in the world!",
+        },
+        'courses': {
```
roxanariza, line 990: Invalid key. Should remain ''course' like on drumeo/pianote

src/contentMetaData.js
```diff
@@ -0,0 +1,1154 @@
...
+        'instructor': {
+            description: "Your drumming journey is unique. You need personalized coaching that helps you reach your goals. Learn from some of the best drummers in the world!",
+        },
+        'course': {
+            name: 'Courses',
+            icon: 'icon-courses',
+            description: "Tackle your next drumming goal with bite-sized courses from many of the world's best drummers.",
+            allowableFilters: ['difficulty', 'genre', 'essential', 'theory', 'lifestyle', 'creativity'],
+            sortBy: '-published_on',
+            tabs: [
```
roxanariza, line 183: In general the tabs are the same for content types per each brand; there are only some situations when the tabs can be different. I think that it's a good idea to move tabs in general, if are identical.

roxanariza (changes requested):

src/contentMetaData.js
adrian-musora: I've moved common course information to common, good catch on the 'course' vs 'courses'

PR #67, opened 2024-09-30
BE 338 coach lessons

roxanariza:
Support for coach lessons filters:
- define metadata for coach lessons
- update filters to support filter by lesson type
- update filter options to support lesson type


src/services/sanity.js
```diff
@@ -599,25 +600,44 @@ export async function fetchAllFilterOptions(
...
+    if (coachId) {
+        // Coach-specific filtering
+        commonFilter = `brand == '${brand}' && references(*[_type=='instructor' && railcontent_id == ${coachId}]._id) ${includedFieldsFilter ? includedFieldsFilter : ''}`;
+    } else {
+        // Regular content filtering
+        commonFilter = `_type == '${contentType}' && brand == "${brand}"${style ? ` && '${style}' in genre[]->name` : ''}${artist ? ` && artist->name == '${artist}'` : ''} ${progressFilter} ${includedFieldsFilter ? includedFieldsFilter : ''}`;
+    }
+
+    // Determine metadata and allowable filters (handle coach lessons if coachId exists)
+    const metaData = processMetadata(brand, coachId ? 'coach-lessons' : contentType, true);
```
BrandonKerr, line 624: Could the contentType be something other than 'coach-lessons'? If not, we could drop this change and it might be a good idea to add some sort of validation at the first step.
Looking at the test, I see you can just set it to `''` to get around it. It feels a bit off to be able to set it to something that will be ignored when the coachId is set.

src/services/sanity.js
```diff
@@ -599,25 +600,44 @@ export async function fetchAllFilterOptions(
     artist,
     contentType,
     term,
-    progressIds = undefined
+    progressIds = undefined,
+    coachId = undefined, // New parameter for coach ID
```
BrandonKerr, line 604: This is just missing from the JS doc for the function

BrandonKerr (approved):
LGTM, just a thought on how it might impact the content type handling.

PR #69, opened 2024-10-02
BE-342 fetch coach lessons filters support

roxanariza:
Add filters support on fetchCoachLessons; update docs, create new test


src/services/sanity.js
```diff
@@ -1084,6 +1084,12 @@ export async function fetchChallengeOverview(id) {
  * Fetch the data needed for the coach screen.
  * @param {string} id - The Railcontent ID of the coach
```
BrandonKerr, line 1085: Small piece and not your change (I just noticed because of your other JSDoc update), but `brand` is missing in the JSDoc.
```suggestion
 * @param {string} brand - The brand for which to fetch coach lessons
 * @param {string} id - The Railcontent ID of the coach
```

BrandonKerr (approved):
Looks good

PR #74, opened 2024-10-08
Adds railcontent_id to group by fetch all.

DidierF:


src/services/sanity.js
```diff
@@ -518,6 +518,7 @@ export async function fetchAll(brand, type, {
         const lessonsFilter = `_type == '${type}' && brand == '${brand}' && ^._id == ${groupBy}._ref ${searchFilter} ${includedFieldsFilter} ${progressFilter}`;
         entityFieldsString = `
                 'id': _id,
+                railcontent_id,
```
roxanariza, line 521: In this case (grouped by artist), the railcontent_id will always be null.

src/services/sanity.js
```diff
@@ -534,6 +535,7 @@ export async function fetchAll(brand, type, {
         const lessonsFilter = `brand == '${brand}' && ^._id in ${groupBy}[]._ref ${typeFilter} ${searchFilter} ${includedFieldsFilter} ${progressFilter}`;
         entityFieldsString = `
                 'id': _id,
+                railcontent_id,
```
roxanariza, line 538: In this case (grouped by genre), the railcontent_id will always be null.

roxanariza (approved):
Your updates look good!
For group_by(artist/genre) the railcontent_id will be null always, but I supposed that you want the response structure to be the same; so I am fine with this.

I am not sure why we need "instructor" and "coaches" on fetchLessonContent. Are duplicate data, but with different fields

src/services/sanity.js
DidierF: In some cases (workouts, for example), group by instructor is using this case, so railcontent_id is the instructor's railcontent_id. 🤔

roxanariza: Yes, when we have group by instructor the railcontent_id will be set.

PR #86, opened 2024-10-22
User Progress context

robwillems:
Accompanying logic for MWP PR
https://github.com/railroadmedia/musora-web-platform/pull/1529


src/services/sanity.js
```diff
@@ -1374,6 +1374,77 @@ export async function fetchGenreLessons(brand, name, contentType, {
...
+                  railcontent_id,               
+            } 
+          }
+        }
+      }
+    }`;
+    let response = await fetchSanity(query, false, {processNeedAccess: false});
+    if (!response) return null;
+    let currentLevel = response;
+    for (let i = 0; i < 4; i++) {
```
BrandonKerr, line 1396: I wonder if this `4` should be an optional param for the function, in case of different content types or pages requiring different levels of hierarchy. (We'd also need to build the query dynamically based on that number to make sure we have the right number of levels).

BrandonKerr: I also remembered that Roxana has done more work on hierarchy lately and I wonder if we could just directly call something from the piece of content in Sanity?

src/services/sanity.js
```diff
@@ -1374,6 +1374,77 @@ export async function fetchGenreLessons(brand, name, contentType, {
...
+        } else {
+            return currentLevel['railcontent_id'];
+        }
+    }
+    return null;
+}
+
+export async function fetchHierarchy(railcontentId) {
+    let topLevelId = await fetchTopLevelParentId(railcontentId);
+    const query = `*[railcontent_id == ${topLevelId}]{
```
BrandonKerr, line 1408: Same thought on if we can get this data directly from the Sanity data, and if not, if it should be configurable for how many levels we go to.

BrandonKerr (approved):
Looks good

src/services/sanity.js
robwillems: This wasn't trivial so i did not do it.  Also the intention here is to have the full hierarchy pulled down in one shot so we can make use of the caching for lots of different ids.

src/services/sanity.js
robwillems: This is meant to pull the top entity, so I don't know if having a depth here makes sense.  There could be a function to pull to a certain depths but if you don't know where your at I'm not sure what the use cases for that would actually be.  If we need it in the future we could revamp these functions to do that.

PR #91, opened 2024-11-04
Optimizations for User Permissions

robwillems:
Added some optimizations for user permissions.

Main issue is every sanity request we make a separate call after the sanity request to get the users permissions.  I changed this to triggered at the same time as the sanity request using Promise.all.
I also made it so only one promise was created so we don't create duplicate requests simultaneously (I'm not an expect with promises so I'll need some feedback for FE/MA whether or not this technique will actually work.

### DIFF
--- a/src/services/railcontent.js
+++ b/src/services/railcontent.js
@@ -239,7 +239,7 @@ export async function fetchContentPageUserData(contentId) {
 export async function fetchUserPermissions() {
     let url = `/content/user_data_permissions`;
     // in the case of an unauthorized user, we return empty permissions
-    return fetchHandler(url, 'get') ?? [];
+    return await fetchHandler(url, 'get') ?? [];
 }
 
 async function fetchDataHandler(url, dataVersion, method = "get") {
--- a/src/services/railcontent.js
+++ b/src/services/railcontent.js
@@ -238,7 +238,7 @@ export async function fetchContentPageUserData(contentId) {
 }
 
 export async function fetchUserPermissionsData() {
-    let url = `/content/user_data_permissions`;
+    let url = `/content/user/permissions`;
     // in the case of an unauthorized user, we return empty permissions
     return await fetchHandler(url, 'get') ?? [];
 }
--- a/src/services/railcontent.js
+++ b/src/services/railcontent.js
@@ -16,7 +16,8 @@ const excludeFromGeneratedIndex = [
     'postRecordWatchSession',
     'postContentStarted',
     'postContentCompleted',
-    'postContentReset'
+    'postContentReset',
+    'fetchUserPermissionsData'
 ];
 
 
--- a/src/services/railcontent.js
+++ b/src/services/railcontent.js
@@ -236,7 +236,7 @@ export async function fetchContentPageUserData(contentId) {
     }
 }
 
-export async function fetchUserPermissions() {
+export async function fetchUserPermissionsData() {
     let url = `/content/user_data_permissions`;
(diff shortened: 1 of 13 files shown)


src/services/userPermissions.js
```diff
@@ -0,0 +1,12 @@
+import {
+    fetchUserPermissionsData
+} from "./railcontent";
+
+let userPermissionsPromise = null;
+
+export async function fetchUserPermissions() {
+    if (!userPermissionsPromise) {
```
robwillems, line 8: My intention here is to ensure there is only one execution of this promise per request.  My understanding is on every request the dom is rebuilt in the browser, so this variable (userPermissionsPromise) will be cleared on every refresh.  @miguelfondeur is this correct or I'm I missing something in my understanding here.

I am not sure how this will behave on Mobile  @csillaj is this behavior similar?  Is there a DOM that's rebuilt every request will this persist somehow?

If not is there some other way to ensure we only call a promise once every request on both FE/MA?

csillaj: @robwillems React Native doesn't have a DOM in the same way as web has, it has a Virtual DOM. Every time a component changes, RN doesn't re-render the whole DOM, it searches where the change occurred and calculates the minimal changes needed. Also in RN DOM elements are translated to platform-specific (android, ios) UI components.
Something else I also might have to mention is that in the app we don't (can't) use this library as a script embedded in html, we import each method we are using directly from the library in our services.
So if this fetchUserPermissions is called only once in a method, than when we import and call the method (e.g.: fetchAll) fetchUserPermissions will only be called once.

BrandonKerr (commented):
I like the idea you have, but am way beyond my JS (and especially RN) knowledge to really weigh in on this. From Csilla's comment, it sounds like this won't actually work as intended. It makes me wonder if this is something that could go into local storage (or whatever mechanism that works with both web and RN): On request start, clear then cache the permissions. That would make sure it gets set once and then used for the lifetime of the request.

src/services/userPermissions.js
robwillems: Thanks, @csillaj. That's what I needed to know.  I'll come up with something else for this.

src/services/railcontent.js
```diff
@@ -236,10 +237,10 @@ export async function fetchContentPageUserData(contentId) {
     }
 }
 
-export async function fetchUserPermissions() {
-    let url = `/content/user_data_permissions`;
+export async function fetchUserPermissionsData() {
+    let url = `/content/user/permissions`;
```
adrian-musora, line 241: Is there a MWP PR for this that I'm missing? I couldn't find this route.

test/initializeTests.js
```diff
@@ -0,0 +1,23 @@
...
+            useCachedAPI: process.env.SANITY_USE_CACHED_API === 'true' || true,
+            version: '2021-06-07',
+            debug: process.env.DEBUG === 'true' || false,
+            useDummyRailContentMethods: true,
+        },
+        localStorage: new LocalStorageMock()
+    };
+    initializeService(config);
+
+    let mock = jest.spyOn(railContentModule, 'fetchUserPermissionsData');
```
adrian-musora, line 20: this is so nice.

adrian-musora (approved):
Approved, needs the MWP endpoint update before merging but otherwise good.

src/services/railcontent.js
robwillems: I just pushed it because it was very straightforward.
https://web-staging-one.musora.com/content/user/permissions

PR #98, opened 2024-11-13
deterministic uuid on recordWatchSession

yalungtang:


BrandonKerr (commented):
If I'm not mistaken, wouldn't this create a new UUID every time? As an example workflow, we could have 
1. the user starts a new session (sessionId is null)
 i. this means we return uuidv4 (e.g. `ae94f6db-51a3-4d39-aec3-0911e3317732`)
 ii. BE gets `ae94f6db-51a3-4d39-aec3-0911e3317732` as the identifier for the session, which doesn't match an existing one and creates a new one
2. the user's watch session is updated, so we pass in the sessionId of `ae94f6db-51a3-4d39-aec3-0911e3317732`
 i.  this means we return a uuidv5 based on that sessionId, which is now `973f6ad3-39b3-508e-b7dd-66ef9a17d8f4`
ii. BE gets `973f6ad3-39b3-508e-b7dd-66ef9a17d8f4` as the identifier for the session, which doesn't match an existing one and creates a new one
3. the user's watch session is updated, so we pass in the sessionId of `973f6ad3-39b3-508e-b7dd-66ef9a17d8f4`
 i. this means we return a uuidv5 based on that sessionId, which is now `f1b1a114-9886-5931-b665-e6d8eaca1fed`
ii. BE gets `f1b1a114-9886-5931-b665-e6d8eaca1fed` as the identifier for the session, which doesn't match an existing one and creates a new one
...

@robwillems is much more familiar with the user's content progress so I'll let him weigh-in, as I could be mistaken on the flow

yalungtang:
> If I'm not mistaken, wouldn't this create a new UUID every time? As an example workflow, we could have
> 
> 1. the user starts a new session (sessionId is null)
>    i. this means we return uuidv4 (e.g. `ae94f6db-51a3-4d39-aec3-0911e3317732`)
>    ii. BE gets `ae94f6db-51a3-4d39-aec3-0911e3317732` as the identifier for the session, which doesn't match an existing one and creates a new one
> 2. the user's watch session is updated, so we pass in the sessionId of `ae94f6db-51a3-4d39-aec3-0911e3317732`
>    i.  this means we return a uuidv5 based on that sessionId, which is now `973f6ad3-39b3-508e-b7dd-66ef9a17d8f4`
>    ii. BE gets `973f6ad3-39b3-508e-b7dd-66ef9a17d8f4` as the identifier for the session, which doesn't match an existing one and creates a new one
> 3. the user's watch session is updated, so we pass in the sessionId of `973f6ad3-39b3-508e-b7dd-66ef9a17d8f4`
>    i. this means we return a uuidv5 based on that sessionId, which is now `f1b1a114-9886-5931-b665-e6d8eaca1fed`
>    ii. BE gets `f1b1a114-9886-5931-b665-e6d8eaca1fed` as the identifier for the session, which doesn't match an existing one and creates a new one
>    ...
> 
> @robwillems is much more familiar with the user's content progress so I'll let him weigh-in, as I could be mistaken on the flow

Before calling the tracking function, we get the sessionToken from an element in the DOM, using it's id.
const sessionTokenElement = document.querySelector('#sessionToken');
This element is added in a blade file on every playback experience Ex. lesson.blade.php
<input type="hidden" id="sessionToken" value="{{ railtracker_session_token() }}">
The format of this token is not in the uuid format, which means, the API fails if you send the current token id. Most probably, if the user is logged in we will have this railtracker session token in place for every call, and the uuidv4 function most probably will never be called, and uuidv5 would give a deterministic representation of that token in uuid format. The previous implementation of uuidv4 will give us a new id for every call, that's the reason why I'm creating the PR.
This is my railtracker token when watching one video
eyJpdiI6IkVRaFVkNTlNYVZOUmVxbXAzNlc2dmc9PSIsInZhbHVlIjoibVZpb2REbmFHalVEU0QydlEwS0w1RythRmd6TGp0aFpKaGZxWTJzbzZwWkZhQ1ZHd0JobW44TnczbUplb1lnM0lGbXlRQ2poUTFxSGhXQUs2czJuT0E9PSIsIm1hYyI6ImVmNmQ4ODZjYWVlZTg3OWFlYWY4ZGM1MzMyYzY5N2U0ODE4ZTgyOWYzYTQ5YmYyZTdjNGVmZDM2ZGU3NmFiYTUiLCJ0YWciOiIifQ==
And this is the token on a different video, it's the same
eyJpdiI6IjZxNVR6R0tzeHU0MGNwUDBNSFpKNEE9PSIsInZhbHVlIjoidHpQNGFSY1JMRzIrVmF1YlhVUWc2MnFRM3JPUkNydUVINHpxaStWcDdkSWV0WkE4LzB3dXFCc1htQU9ydFVLdXlZam9HSmJOSHhsUy8zazlSOXlNQ2c9PSIsIm1hYyI6IjA4OWE5MzE4ZDVhYjc1YmM2YTFmZTIzZDE4YjhiYTllNzkxNjZmZDg1MzFlMTg0OTBlMmE1MDgzODIwMTdmNWEiLCJ0YWciOiIifQ==
As long as we use the same namespace to transform the token the result will be the same

yalungtang:
I discussed with @robwillems that we don't need an id for the duration of the session throughout the app, we just need to keep the same id for the duration of the watch session. So this is not needed at all.

PR #103, opened 2024-11-20
Basic users should see the songs catalog behind membership upgrade modal

roxanariza:
pullSongsContent - Allows pull songs content - user->admin/membership_level

Basic users should see the songs catalog behind membership upgrade modal


src/services/sanity.js
```diff
@@ -511,6 +511,7 @@ export async function fetchAll(brand, type, {
     const start = (page - 1) * limit;
     const end = start + limit;
     let bypassStatusAndPublishedValidation = (type == 'instructor');
+    let allowsPullSongsContent = (type == 'song');
```
adrian-musora, line 514: technically this isn't just song content. But is anything with the musoraPlus membership.

The logic for this on the BE is in `SetContentPermissions.php`
```
 if (!user()->isAdmin() && user()->isABasicMember()) {
                ContentRepository::$allowsPullSongsContent = true;
            }
            ```

roxanariza: @adrian-musora - In the middleware, we set ContentRepository::$allowsPullSongsContent to true, but many controller methods override it by setting it to false. I've identified that we only need ContentRepository::$allowsPullSongsContent to be true for the Songs catalog page, where we display songs alongside the Upgrade Membership modal. Please let me know if I missed any other pages or sections where this is required.

adrian-musora: Based on my search and memory, only the homepage getNewContents sets this filter to false.
IIRC: The desired behaviour was also for lesson pages, recommended (though for that we hide the songs tab). where we.
I've put a message in chat to confirm, but I don't recall it just being songs catalogue page when I implemented the filter.

roxanariza: Updated the flag based on the comment and discussion

adrian-musora (approved):

PR #106, opened 2024-11-22
TP-322: Focus Field for All Coaches shows an [Object Object]

adrian-musora:


src/contentTypeConfig.js
```diff
@@ -345,12 +345,12 @@ let contentTypeConfig = {
             '"coach_card_image": coach_card_image.asset->url',
             '"coach_featured_image": coach_featured_image.asset->url',
             '"coach_top_banner_image": coach_top_banner_image.asset->url',
-            'focus',
+            '"focus": focus[]->name',
             'focus_text',
             'forum_thread_id',
-            'long_bio',
+            '"long_bio": long_bio[0].children[0].text,',
             'name',
```
csillaj, line 352: @adrian-musora this groq fails now after your changes, because you have two commas between long_bio and name

PR #107, opened 2024-11-26
Add url to contentMetadata

roxanariza:
- Provide custom url for some Drumeo shows that have different type in Sanity
- Update processMetadata method to return a new field: url. The new field is the custom url if it was defined for selected type or the type


BrandonKerr (commented):
I like the general idea but I think we'll need to add the url to every entry, won't we? At least for all the shows that will be using that value - e.g. there's no url for the _on-the-road_ entry, so the Vue component won't know what to put in there.
I also think that the MWP side of it already has the `/` between the base url and the type portion, so we might need to do `url: 'gear-guides'` instead of `url: '/gear-guides'`, etc.

PR #113, opened 2024-12-02
Fix access to documents without permisions set

roxanariza:
Fix access to documents without permisions set


BrandonKerr (approved):
Looks great, but I think a test might need to be updated since it's failing

roxanariza:
> Looks great, but I think a test might need to be updated since it's failing

Yeah,  it is failing from yesterday, but I will investigate what's wrong with the published_on restrictions

PR #114, opened 2024-12-02
Fix scheduled contents display on catalogues:

roxanariza:
- add getFutureScheduledContentsOnly on catalogues for admin and members
 - fix statuses when Scheduled exists in availableContentStatuses


BrandonKerr (approved):
Looks great!

PR #117, opened 2024-12-04
Next lesson data

BrandonKerr:
[TP-402](https://musora.atlassian.net/browse/TP-402)
Associated [MWP PR](https://github.com/railroadmedia/musora-web-platform/pull/1703)

Adds a `jumpToContinueContent` function to retrieve the next document from Sanity for the given parent railcontent ID, as well as a `fetchNextContentDataForParent` function to get the next content's ID and type from MWP. This should work as a replacement for Musora API's jumpToContinueContent.

[TP-402]: https://musora.atlassian.net/browse/TP-402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/railcontent.js
```diff
@@ -239,6 +239,36 @@ export async function fetchContentPageUserData(contentId) {
...
 
+/**
+ * Fetches the ID and Type of the piece of content that would be the next one for the user
+ *
+ * @param {int} contentId - The id of the parent (method, level, or course) piece of content.
+ * @returns {Promise<Object|null>} - Returns and Object with the id and type of the next piece of content if found, otherwise null.
+ */
+export async function fetchNextContentDataForParent(contentId) {
+    let url = `/content/${contentId}/next/${globalConfig.railcontentConfig.userId}`;
+    const headers = {
```
adrian-musora, line 250: To avoid code duplication (the other methods should also use the fetch handler)

```suggestion
       return fetchHandler(url); 
```

src/services/railcontent.js
```diff
@@ -239,6 +239,36 @@ export async function fetchContentPageUserData(contentId) {
...
+            // console.log('fetchNextContentDataForParent', result);
+            return result.next;
+        } else {
+            console.log('fetchNextContentDataForParent result not json');
+            return null;
+        }
+    } catch (error) {
+        console.error('Fetch error:', error);
+        return null;
+    }
```
adrian-musora, line 268: to avoid the duplication. The above functions should also use this, but here we are.
```suggestion
    return fetchHandler(url);
```

adrian-musora (approved):

csillaj:
@BrandonKerr could you deploy `/content/${contentId}/next/` to web-staging/app-staging when merging this?

src/services/railcontent.js
BrandonKerr: That is a much better idea. Sounds like a nice and hopefully pretty small refactor for next year

BrandonKerr:
> @BrandonKerr could you deploy `/content/${contentId}/next/` to web-staging/app-staging when merging this?

@csillaj the MWP side of this is up on web-staging-one now as well

PR #121, opened 2024-12-05
Add accept json to railcontent requests

BrandonKerr:
Our requests to railcontent were missing this header, so Laravel would send back HTML responses (e.g. the 404 page when something wasn't found). This will ensure a proper JSON response is sent. I also ran the build-index and docs, so there are some additional changes beyond what I actually did.


docs/sanity.js.html
```diff
@@ -142,6 +142,11 @@ <h1 class="page-title">sanity.js</h1>
     return fetchSanity(query, true, {processNeedAccess: false});
 }
 
+export async function fetchPlayAlongsCount(brand) {
```
adrian-musora, line 145: Did you branch from a WIP branch? I don't think these changes should be here.

adrian-musora (approved):

PR #128, opened 2024-12-10
Change sanity to use posts

robwillems:


src/services/sanity.js
```diff
@@ -1769,15 +1769,22 @@ export async function fetchSanity(query,
     const perspective = globalConfig.sanityConfig.perspective ?? 'published';
     const encodedQuery = encodeURIComponent(query);
```
BrandonKerr, line 1770: I think we could drop this now
```suggestion
```

BrandonKerr (approved):
Awesome, thanks for the quick fix!

PR #133, opened 2024-12-13
Be 400 live and chat

roxanariza:
New method for chat an live event


src/services/sanity.js
```diff
@@ -28,7 +28,7 @@ import {
     fetchCompletedChallenges,
     fetchCurrentSongComplete,
     fetchOwnedChallenges,
-    fetchNextContentDataForParent
+    fetchNextContentDataForParent, fetchHandler
```
adrian-musora, line 31:
```suggestion
    fetchNextContentDataForParent,
    fetchHandler,
```

adrian-musora (approved):

PR #137, opened 2024-12-17
Dev/rwillems20241216 tp 581

robwillems:
Handle TP-581, issues with switching user.
Changed localstorage keys to include userId

### DIFF
--- a/src/services/dataContext.js
+++ b/src/services/dataContext.js
@@ -24,7 +24,7 @@ export class DataContext {
     }
 
     getLocalStorageKey() {
-        const userId = globalConfig.railcontentConfig.userId ?? 0;
+        const userId = globalConfig.railcontentConfig?.userId ?? 0;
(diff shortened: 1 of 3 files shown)


csillaj (approved):

csillaj:
@Rob I approved it but then after testing it more I've found that the issue still persists
even if after logging in with the second user, the key of the storage contains the new user's id, the data in the cache is the same

csillaj:
Looking into it more:
when I log in with the second user:
`let lastUpdated = globalConfig.isMA ? await cache.getItem(this.getLocalStorageLastUpdatedKey()) : cache.getItem(this.getLocalStorageLastUpdatedKey());` -> here lastUpdated is null, so false is returned by shouldVerifyServerVerions
so in getDataPromise, this if is not entered  `if ((!this.context) || shouldVerify) `

PR #144, opened 2024-12-19
Add status filter for fetchRelatedSongs

BrandonKerr:
[TP-775](https://musora.atlassian.net/browse/TP-775)

This adds the necessary status filtering to keep unpublished content from showing up in the results of fetching related songs.

A better solution is to use the filter builder, but this was a lot faster to implement.

[TP-775]: https://musora.atlassian.net/browse/TP-775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robwillems (approved):
Looks good enough for now.  Would be nice to integrate filterbuilder eventually, but not a big deal if it never happens.

PR #147, opened 2024-12-27
added breadcrumbs to method related functions

yalungtang:


adrian-musora (approved):
Works great. I tested locally by checking the debug sanity results in console for drumeo method.

Small improvement, The code is repeated in three places, you could add something similar to the `descriptionField` property.

src/services/sanity.js
```diff
@@ -928,6 +928,12 @@ export async function fetchParentByRailContentId(railcontentId) {
  */
 export async function fetchMethods(brand) {
     const query = `*[_type == 'learning-path' && brand == '${brand}'] {
+    parent_content_data,
+    "breadcrumbs_data": parent_content_data[] {
+        "id": id,
+        "title": *[railcontent_id == ^.id][0].title,
```
robwillems, line 934: I'm not a sanity expert, but can you just go ^.title to get the top level data?  This query does not seem like it would be very efficient unless it does some magic internally.  Does this slow down the query at all?

robwillems (commented):
Not sure about the performance of this.

src/services/sanity.js
robwillems: Nvm, parent_content_data is not a reference so this is probably the best we can do at the moment.

robwillems (approved):
Nvm, performance is best we can do at the moment.

PR #165, opened 2025-02-04
BEH-212: Song Tutorials

DidierF:
- Added a few fields needed for [BEH-212](https://musora.atlassian.net/browse/BEH-212).
- Moved custom fields after default because lessons was being overwritten by the default without the extra fields.

Please let me know if this could cause issues and if I should test anything specific.

[BEH-212]: https://musora.atlassian.net/browse/BEH-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/sanity.js
```diff
@@ -342,7 +342,7 @@ export async function fetchScheduledReleases(brand, { page = 1, limit = 10 }) {
 export async function fetchByRailContentId(id, contentType) {
   const fields = getFieldsForContentType(contentType)
   const childrenFilter = await new FilterBuilder(``, { isChildrenFilter: true }).buildFilter()
-  const entityFieldsString = ` ${fields}
+  const entityFieldsString = `
```
adrian-musora, line 345: the `child_count` and `lessons` objects include the `childFilter` query which restricts document based on `status` `published_on` and `permissions`. Not including this will show restricted content to unauthorized users, or show draft content.

Also we have `child_count` and `lesson_count` we should consolidate those two fields and remove the duplication of information. Lesson_count will be incorrect for songs that have some restricted tutorial-parts.

There are two options that I see to handle this.
1. Add the new fields to the lessons object in `fetchByRailContent`
2. refactor `fetchByRailContentId` to have get a `childsField` string that is generated the same way that `getFieldsForContentType` makes the parent.

I don't have a strong preference, 2 will be more extendable in the future, but is a bit more work now. 
Regardless the song-tutorial entry in `contentTypeConfig` should be cleaned up to match based on 1 or 2.

DidierF: That makes sense, I'll go with 2.

src/contentTypeConfig.js
```diff
@@ -111,6 +118,19 @@ export const coachLessonsTypes = [
   'workout',
 ]
 
+export const childContentTypeConfig = {
+  'song-tutorial': [
+    `"id": railcontent_id`,
+    `title`,
+    `"image": thumbnail.asset->url`,
+    `"instructors": instructor[]->name`,
+    `length_in_seconds`,
```
adrian-musora, line 127: The method concats the two arrays so we can remove these ones.
```suggestion
```

adrian-musora (approved):
I didn't test this for functionality, so trusting you the data ends up correct :)

Looks good, thank you.

PR #166, opened 2025-02-07
Tabs + Consts refactor

adrian-musora:


src/contentMetaData.js
```diff
@@ -1,5 +1,45 @@
 // Metadata is taken from the 'common' element and then merged with the <brand> metadata.
 // Brand values are prioritized and will override the same property in the 'common' element.
+
+const PROGRESS_NAMES = ['All','In Progress', 'Complete','Not Started']
```
adrian-musora, line 4: @roxanariza I changed the text from `In progress` to `In Progress` because it made more sense grammatically. Do you know if this casing was intentional?

roxanariza: @adrian-musora - Was not intentional

roxanariza (approved):
Most of the configurations will be deleted, but I like the idea. 
I approve the PR

PR #173, opened 2025-02-26
BE-308 - add comments endpoints

adrian-musora:


src/services/railcontent.js
```diff
@@ -1198,6 +1198,43 @@ export async function setStudentViewForUser(userId, enable) {
   return await patchDataHandler(url, data)
 }
 
+
+/**
+ * Fetch the top comment for a given content
+ *
+ * @param {int} railcontentId - The railcontent id to fetch.
+ * @returns {Promise<Object|null>} - A promise that resolves to an comment object
```
alesevero, line 1206: I'd add the types here as well. For example [here](https://github.com/railroadmedia/musora-content-services/blob/347990cd565859e1c258bf2574f81e4b79462fc7/src/services/user/sessions.js#L24)

I've created a types.js file within `services/user/` in #175 so we can add for this one later as well when we move these services around

alesevero (approved):
Looks good!

PR #182, opened 2025-03-10
MU2-248 streak module

roxanariza:
New method getUserActivityStats that returns dummy data


alesevero (approved):
Looks good. We might need to update schemas and stuff after our meeting tomorrow but ti is a good start!

PR #192, opened 2025-03-17
Store all user_practice in dataContext; new method to calculate curre…

roxanariza:
MU2-346; MU2-347; MU2-348

- store all user practice in dataContext; 
- new method to calculate current week stats
- new method to calculate monthly stats

Related with PR: https://github.com/railroadmedia/musora-platform-backend/pull/25


src/services/userActivity.js
```diff
@@ -23,10 +19,210 @@ const userActivityStats = {
   streakMessage: "That's 8 weeks in a row! Way to keep your streak going.",
```
adrian-musora, line 19: Are these intended as examples? I think they should be removed as they're not used here that I could tell.

adrian-musora (approved):
Looks good to me.
I made a few changes to the tests so they can run in parallel and added the consoleLog utility method.

src/services/userActivity.js
```diff
@@ -23,10 +19,210 @@ const userActivityStats = {
...
+
+  let dailyStats = [];
+  for (let i = 0; i < daysInMonth; i++) {
+    let day = new Date(startOfMonth);
+    day.setDate(startOfMonth.getDate() + i);
+    let dayKey = day.toISOString().split('T')[0];
+
+    // Check if the user has activity for the day, default to 0 if undefined
+    let dayActivity = practices[dayKey] ?? null;
+    let isActive = dayKey === today.toISOString().split('T')[0];
```
adrian-musora, line 173: Small optimization
```suggestion
  const todayKey = today.toISOString().split('T')[0]
  for (let i = 0; i < daysInMonth; i++) {
    let day = new Date(startOfMonth);
    day.setDate(startOfMonth.getDate() + i);
    let dayKey = day.toISOString().split('T')[0];

    // Check if the user has activity for the day, default to 0 if undefined
    let dayActivity = practices[dayKey] ?? null;
    let isActive = dayKey === todayKey;
```

adrian-musora:
Code should be prettied

src/services/userActivity.js
adrian-musora, line 1: File should be prettied. I didn't want to do it to avoid adding a bunch of changes for you sift through.

adrian-musora:
The RFC doc specifies a bunch of extra tests, do you want me to write these? or make a ticket for later? or leave it?
```Test Cases:
Transition from one day to the next
Transition from not active in current day to active
Transition from one week to the next
Transition from not active in current week to active
Timezone testing
```

src/services/userActivity.js
```diff
@@ -2,31 +2,228 @@
...
+    if (prevDay === null || isNextDay(prevDay, dayKey)) {
+      dailyStreak++
+      longestDailyStreak = Math.max(longestDailyStreak, dailyStreak)
+    } else {
+      dailyStreak = 1 // Reset streak if there's a gap
+    }
+    prevDay = dayKey
+  })
+
+  currentDailyStreak = longestDailyStreak
```
adrian-musora, line 83: This is not the current daily streak, this is only valid if the streak is current ongoing.
Should we be returning all three values?
`longestDailyStreak`, `currentDailyStreak`, and `currentWeeklyStreak`

src/services/userActivity.js
```diff
@@ -2,31 +2,228 @@
...
+      if (currentWeekActivity) {
+        weeklyStreak++
+        currentWeekActivity = false
+      }
+      prevWeek = weekNumber
+    }
+
+    if (practices[dayKey]) {
+      currentWeekActivity = true
+    }
```
adrian-musora, line 108: I think this check needs to go before the `else if (weekNumber !== prevWeek` clause, as if there is a skip of more than one week between dates stored it `currentWeekActivity` won't be reflected.

src/services/userActivity.js
```diff
@@ -2,31 +2,228 @@
...
+
+export async function getUserMonthlyStats(year = new Date().getFullYear(), month = new Date().getMonth(), day = 1) {
+  let data = await userActivityContext.getData()
+  let practices = data?.[DATA_KEY_PRACTICES] ?? {}
+
+  // Get the first day of the specified month and the number of days in that month
+  let firstDayOfMonth = new Date(year, month, 1)
+  let today = new Date()
+
+  let startOfMonth = getMonday(firstDayOfMonth)
```
adrian-musora, line 156: Why is this shifted to the Monday? Shouldn't we be calculated to the first day of the month?

src/services/userActivity.js
```diff
@@ -2,31 +2,228 @@
...
-};
+  streakMessage: 'That\'s 8 weeks in a row! Way to keep your streak going.',
+}
+
+const DATA_KEY_PRACTICES = 'practices'
+const DATA_KEY_LAST_UPDATED_TIME = 'u'
+
+const DAYS = ['M', 'T', 'W', 'T', 'F', 'S', 'S']
 
 export async function getUserActivityStats(brand) {
```
alesevero, line 28: It would be valuable for FA and MA teams for have some documentation on the functions being exposed from MCS.

src/services/userActivity.js
roxanariza: Yes, they are the hardcoded data provided to MA/FE teams for streaks. will be removed

src/services/userActivity.js
```diff
@@ -2,31 +2,228 @@
...
+  const todayKey = today.toISOString().split('T')[0]
+  for (let i = 0; i < 7; i++) {
+    let day = new Date(startOfWeek)
+    day.setDate(startOfWeek.getDate() + i)
+    let dayKey = day.toISOString().split('T')[0]
+
+    let dayActivity = practices[dayKey] ?? null
+    let isActive = dayKey === todayKey
+    let type = (dayActivity !== null ? 'tracked' : (isActive ? 'active' : 'none'))
+    dailyStats.push({ key: i, label: DAYS[i], isActive, inStreak: dayActivity !== null, type, myData: dayKey })
```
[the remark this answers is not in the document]
roxanariza: myData was inserted only for some test   Thanks!

src/services/userActivity.js
roxanariza: The FE and MA teams will have the documentation

src/services/userActivity.js
roxanariza: Only currentDailyStreak and currentWeeklyStreak should be returned

src/services/userActivity.js
```diff
@@ -2,31 +2,228 @@
...
+  )
+}
+
+function getStreaksAndMessage(practices)
+{
+  let { currentDailyStreak, currentWeeklyStreak } = calculateStreaks(practices)
+
+  let streakMessage = currentWeeklyStreak > 1
+    ? `That's ${currentWeeklyStreak} weeks in a row! Keep going!`
+    : ''
```
adrian-musora, line 218: I recall that we wanted the daily streak to be shown if exists then the weekly streak? I couldn't find the document on it though. something like this?
```suggestion
  let streakMessage = currentDailyStreak > 0 ? `That's ${currentDailyStreak} Days in a row! Keep going!` :
    currentWeeklyStreak > 1 ? `That's ${currentWeeklyStreak} weeks in a row! Keep going!` : ''
```

roxanariza: In my opinion, weekly streaks are more important than daily streaks, but I will put a question. Anyway, I will provide a different response depending on what method(monthly/weekly) is called. For weekly, I need to provide only the streak message and the dailyActive Stats and for monthly, I will provide: dailyActiveStats, weeklyActiveStats, total practice duration, current weekly streak and current daily streak

src/services/userActivity.js
adrian-musora: Ah, I see. So the domain is the (monday before or 1st) TO (last day of the month or the next sunday)

So `daysInMonth` needs to be adjusted accordingly

PR #194, opened 2025-03-18
Remove unused functions found during BE cleanup

robwillems:


Joels-code (approved):
Doesn't look like we use these in V2

PR #199, opened 2025-03-22
BEH-313: Add related lesson endpoints

adrian-musora:
[JIRA](https://musora.atlassian.net/browse/BEH-313)

Design:
- i did one endponit for each module, as different content types will use different modules and I like the SPA style 
- I refactored the testsuites to have slightly better names, and added an admin version for testing content that's blocked behind admin issues for license reasons.
- Single parent types will potentially be used by FEW/MA to determine if the related lessons module should be called.
- Song types, may change names (jam-tracks don't exist yet) but that will be my problem later when I add the type

Testing:
- I did all my testing in the testsuite. 
- I added a test for each endpoint with the provided content type. I did not add a recommended test because we don't have it configured for testing and I can't be bothered right now.
- lessons in this X, were tested by passing an a child and verifying we find related content based on the web_url_path.
- license related information was tested by verifing the items are related by licenes and the content type matches the allow/reject filter for different song types.
- Neither of these tests are exaustive for ordering (which I assumed was fine as previously written)

### DIFF
--- a/src/contentTypeConfig.js
+++ b/src/contentTypeConfig.js
@@ -3,6 +3,12 @@ import {Tabs} from "./contentMetaData.js";
 
 export const AWSUrl = 'https://s3.us-east-1.amazonaws.com/musora-web-platform'
 export const CloudFrontURl = 'https://d3fzm1tzeyr5n3.cloudfront.net'
+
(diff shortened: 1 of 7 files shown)


Meowbox64 (approved):
lgtm

PR #200, opened 2025-03-23
Content like endpoint updates

peterminhlee:


src/services/railcontent.js
```diff
@@ -343,17 +343,17 @@ export async function fetchHandler(url, method = 'get', dataVersion = null, body
 }
 
 export async function fetchUserLikes(currentVersion) {
-  let url = `/content/user/likes/all`
+  let url = `/content/v1/user/likes/all`
```
adrian-musora, line 346: Routes go through a different path now. 
Was this tested?
```suggestion
  let url = `/content/v1/user/likes`
```

src/services/railcontent.js
```diff
@@ -343,17 +343,17 @@ export async function fetchHandler(url, method = 'get', dataVersion = null, body
...
   return fetchDataHandler(url, currentVersion)
 }
 
 export async function postContentLiked(contentId) {
-  let url = `/content/user/likes/like/${contentId}`
+  let url = `/content/v1/user/likes/like/${contentId}`
```
adrian-musora, line 351:
```suggestion
  let url = `/content/v1/user/likes/${contentId}`
```

adrian-musora (approved):

PR #201, opened 2025-03-23
Content complete endpoint updates

peterminhlee:


src/services/railcontent.js
```diff
@@ -926,8 +914,8 @@ export async function fetchPlaylistItem(payload) {
   return await fetchHandler(url)
 }
 
-export async function postContentCompleted(contentId) {
-  let url = `/content/user/progress/complete`
+export async function postContentComplete(contentId) {
+  let url = `/content/v1/user/progress/complete`
   return postDataHandler(url, { contentId: contentId })
```
adrian-musora, line 919: Your BE changes include the contentId as part of the url route, not as part of the request body. I'm almost positive you'll need something like this:
```suggestion
  let url = `/content/v1/user/progress/complete/${contentId}`
  return postDataHandler(url)
```

src/index.js
```diff
@@ -96,15 +96,14 @@ import {
...
-	postChallengesCompleteLesson,
 	postChallengesEnroll,
 	postChallengesEnrollmentNotification,
 	postChallengesHideCompletedBanner,
 	postChallengesLeave,
 	postChallengesSetStartDate,
 	postChallengesSoloNotification,
 	postChallengesUnlock,
-	postContentCompleted,
+	postContentComplete,
```
adrian-musora, line 106: These should be excluded from the index files so callers have to go through `contentStatusCompleted`, I can see you've included them in the `excludeFromGeneratedIndex` method. Could you poke into why this isn't working as expected?

adrian-musora (approved):

PR #202, opened 2025-03-25
Add report endpoints 

adrian-musora:
I needed to add the new set of endpoints for user management block/unblock.

Refactored the fetch handler into it's own utility file. It takes a token and baseURl now, so it can be supported by both railcontent.js and recommended.js (or any external JSON API)

I tested this in console, because I don't have a quick way to test through the FEW yet.
To do so, wrote the global config file into console like so. 
token was generated by running `api/user-management-system/v1/sessions` in postmand and pulling the tokens out.
```
globalConfig = {  
railcontentConfig: {  
token: 'pDxpLMl0xtG6hJTQC092Yw5MUKg9UKkSkl2X1ZXx',  
userId: 631736,  
//baseUrl = 'https://dev.musora.com:8443'  
},  
recommendationsConfig: {  
token: '<go grab it from the .env file',  
baseUrl: 'https://MusoraProductDepartment-PWGenerator.hf.space',

}  
}
```

Then dumped the  `fetchHandler` for `recommended.js`, then ran
```await rankItems('drumeo', [410430, 388642, 388199, 411095])```
Then dumped the `fetchHandler` for `railcontent.js` (and the `postHandler`), then ran:
```await createComment(402199, 'im a new comment')```
and then ran the block and unblock commands

### DIFF
--- a/src/services/user/management.js
+++ b/src/services/user/management.js
@@ -12,7 +12,7 @@ import { fetchHandler } from '../railcontent.js'
  */
 const excludeFromGeneratedIndex = []
 
-const baseUrl = `/api/user-management-system/v1`
+const baseUrl = `/api/user-management-system`
 
 /**
  * Block the provided user
@@ -20,7 +20,7 @@ const baseUrl = `/api/user-management-system/v1`
  * @returns {Promise<any|string|null>}
  */
 export async function blockUser(userId) {
-  const url = `${baseUrl}/block/${userId}`
+  const url = `${baseUrl}/v1/block/${userId}`
   return fetchHandler(url, 'post')
 }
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/services/user/management.js [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+
+
+/**
+ * Exported functions that are excluded from index generation.
+ *
+ * @type {string[]}
+ */
+const excludeFromGeneratedIndex = []
+
+const baseUrl = `/api/user-management-system/v1`
```
robwillems, line 15: Minor:
From my understanding of the versioning, we may only update a single route to a new version instead of the entire module at once.  If this is true then having v1 in the baseUrl seems off to me.

robwillems (approved):

src/services/user/management.js
adrian-musora: This is correct, updated.

PR #204, opened 2025-03-26
feat(MU2-357): mock award data

alesevero:
- Add 15 awards to unblock FE/MA
- Fixes sanity.js jsdoc types
- Updates index files
- Updates docs
- Just a bit of formatting


src/services/gamification/gamification.js
```diff
@@ -0,0 +1,4 @@
+/**
+ * @namespace Gamification
```
alesevero, line 2: Looking to the future where awards won't be specifically for challenges

src/services/gamification/types.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * @typedef {Object} Award
```
alesevero, line 2: I think this structure will change for future awards. Will need to wait for Product to present the project

src/services/gamification/awards.js
adrian-musora, line 1: Will this be paginated or do we always serve all data?
If it's paginated we should return a paginated style response.

adrian-musora (approved):
I made some changes to the first 2 elements of the mock data to set them to gold and silver awards and set the values to non-zero. So we can test rendering a bit more properly.

src/services/gamification/awards.js
alesevero: good catch, I forgot about that. Will add it

PR #206, opened 2025-03-26
feat(BEH-341) : part1 - PlayAlong Update: coalesce video from array to flat

adrian-musora:
In the final data structure, the `video` object will be an array of elements. This PR can be finished first so that we can bundle it with other MA changes for when the breaking changes hit playalongs.

You can test this by accessing any lessons data and verifying the video loads.
Any page that uses the lesson 
/pianote/quick-tips/nocturne-in-e-flat-but-make-it-jazz-performance/416936
https://www.musora.com/drumeo/play-alongs/ccr-rock/333131


Meowbox64 (approved):
lgtm

PR #207, opened 2025-03-27
fix(BEH-341) - Add artist to lesson content

adrian-musora:
Artist only has two default fields (name, thumbnail_url)

Testing.
I ran the testsuite locally and verified the output package had a valid image url. Most artists don't have url images, so I updated the railconent_ids for the two tests.

The name of the field is `thumbnail` because I've migrated all the names to match in this [PR](https://github.com/railroadmedia/musora-content-services/pull/205).

### DIFF
--- a/src/services/sanity.js
+++ b/src/services/sanity.js
@@ -59,7 +59,7 @@ export async function fetchSongById(documentId) {
     fields,
     {
       isSingle: true,
-    }
+    },
   )
   return fetchSanity(query, false)
 }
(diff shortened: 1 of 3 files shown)


src/services/sanity.js
```diff
@@ -1260,7 +1261,11 @@ export async function fetchLessonContent(railContentId) {
           instrumentless,
           railcontent_id,
           "id":railcontent_id,
-          slug, artist->,
+          slug,
+          "artist": artist->{
```
adrian-musora, line 1265: here are the non-formatting changes

Meowbox64 (approved):

PR #212, opened 2025-03-31
Add image src build and verify functions

robwillems:
Includes functions for building and verifying image srcs.
Build appends width, height and quality to sanity URLs to reduce bandwidth
Verify checks untransformed sanity routes and direct S3 routes.


calebfavor:
@coderabbitai full review

PR #232, opened 2025-04-15
Added report comment endpoint to MCS

MeenuDhanoa:


src/services/railcontent.js
```diff
@@ -1175,6 +1175,19 @@ export async function editComment(commentId, comment) {
...
+/**
+ * @param {int} commentId
+ * @param {string} issue
+ * @returns {Promise<*|null>}
+ */
+export async function reportComment(commentId, issue) {
+  const url = `/api/content/v1/comments/${commentId}/report`
+  const data = {
+    comment: issue,
+  }
```
adrian-musora, line 1187: I had to update the BE to support this and chose 'issue' as the field name.
```suggestion
  const data = {
    issue: issue,
  }
```

adrian-musora (approved):

PR #251, opened 2025-05-07
Mu2 462 homepage progress rows

roxanariza:
## **Please ignore this PR and review the new [PR](https://github.com/railroadmedia/musora-content-services/pull/262)!!**
I have implemented the homepage progress rows v2 system, a system that is improved and optimized to reduce the number of calls to Sanity.

[MU2-462](https://musora.atlassian.net/browse/MU2-462?atlOrigin=eyJpIjoiM2UzMGJkMGEyZGQyNGU1OWEyN2U1MDExYzgwMWZhNTMiLCJwIjoiaiJ9)

[MPB](https://github.com/railroadmedia/musora-platform-backend/pull/78) 

## Description/Design
New method to provide progress row data sorted by latest interaction
Playlists should be included in progress row cards

More details: https://docs.google.com/document/d/1P_hYpryXGs3kGLn7RM1FFGsYot4bHSiw/edit#heading=h.bosdaabmo79o

I still need to work on:
- shows progress row(need to provide in CTA info for next lesson based on episode number)
- guided courses: provide awards and CTA for Next Lesson in XX:XX(waiting card state)

[MU2-462]: https://musora.atlassian.net/browse/MU2-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
	- Introduced a "Recent Lessons" tab group with Recent All, Recent Incomplete, and Recent Completed tabs for multiple brands.
	- Added the ability to fetch and display progress rows filtered by brand, including recent playlists and content items.
	- Added a new feature to retrieve detailed progress information for multiple content items by their IDs.

- **Bug Fixes**
	- Corrected a lesson type naming issue to ensure consistency in student archive lesson types.


src/contentMetaData.js
```diff
@@ -411,6 +412,14 @@ const contentMetadata = {
...
       ],
     },
+    'recent': {
+      name: 'Recent Lessons',
+      tabs: [
+        Tabs.RecentAll,
+        Tabs.RecentIncomplete,
+        Tabs.RecentCompleted
+      ],
+    },
```
adrian-musora, line 422: This definition should be able to go into commonMetaData and will be accessible by all brands.

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
   return { id: 'recommended', title: 'Recommended For You', items: contents }
 }
 
+export async function getProgressRows({ brand = null, limit = 8 } = {}) {
+  const progressContentsIds = await getAllStartedOrCompleted();
+  const contents = await fetchByRailContentIds(progressContentsIds, 'progress-tracker');
+
+  const filtered = contents.filter(item => item.brand === brand);
```
adrian-musora, line 380: This should happen at the query level in fetchByRailContentIds

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+
+async function buildChildContentMap(hierarchyMap) {
+  const allChildIds = Object.values(hierarchyMap).flatMap(h => Object.keys(h.children || {}));
+  const children = await fetchByRailContentIds(allChildIds);
+  return Object.fromEntries(children.map(child => [child.id, child]));
+}
+
+async function processPlaylistItem(item, progressMap) {
+  const playlist = item.raw;
+  const items = await fetchPlaylistItems(playlist.id);
```
adrian-musora, line 540: This makes a call to the BE for every item (up to `limit`), we could add or update the BE endpoint to filter and include the relevant progress information. Processing could still be done in FE if preferred, but might be easier in SQL

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
   return { id: 'recommended', title: 'Recommended For You', items: contents }
 }
 
+export async function getProgressRows({ brand = null, limit = 8 } = {}) {
+  const progressContentsIds = await getAllStartedOrCompleted();
+  const contents = await fetchByRailContentIds(progressContentsIds, 'progress-tracker');
```
adrian-musora, line 378: This has the potential to be a very large query, as this contains all the content the user has completed or started.
Proposal to filter out the data before calling into sanity for the last ... month? I expect that might need a design change, but it's worth some pushback.

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+const getFormattedType = type => {
+  for (const [key, values] of Object.entries(progressTypesMapping)) {
+    if (values.includes(type)) {
+      return key;
+    }
+  }
+  return null;
+};
+
+async function getCombinedLatestProgressItems(contents, progressMap, playlists, limit = 8) {
```
adrian-musora, line 412: Broadly speaking this will need to be optimized, I'd suggest at another time. We currently make a lot of external calls. And a number of them could be combined or run in parallel. 

1 BE for playlists (which makes 1 sanity call)
1 Sanity call for content

x2 calls Sanity per piece of non-playlist content for heirarchy and childen
1 BE call per playlist item (which also calls sanity)
 
This could be heavily optimized by writing a distinct sanity query to grab all content + children and pre filter for by type.

It's a tricky problem, because we need enough rows to fill out the and don't want to call into sanity for 100 possible entries. I don't have a good answer to this.

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+        raw: p
+      }))
+  );
+}
+
+function mergeAndSortItems(items, limit) {
+  return items
+    .filter(item => typeof item.progressTimestamp === 'number' && item.progressTimestamp > 0)
+    .sort((a, b) => b.progressTimestamp - a.progressTimestamp)
+    .slice(0, limit + 5);
```
adrian-musora, line 517: I couldn't figure out what the limit + 5 was for. I didn't see any filtering after this method call.
Not an issue as far as I can tell, but I like understanding why things are done a particular way.

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+    body: {
+      first_items_thumbnail_url: playlist.first_items_thumbnail_url,
+      title: playlist.name,
+      subtitle: `${playlist.duration_formated} • ${playlist.total_items} items • ${playlist.likes} likes • ${playlist.user.display_name}`
+    },
+    progressTimestamp: item.progressTimestamp,
+    cta: {
+      text: 'Continue',
+      action: playlist.next
+    }
```
adrian-musora, line 601: Some cleanup on this to reduce duplication and dead code.

```suggestion
  let nextId = items[0]?.id ?? null;
  if (!allItemsCompleted) {
    const lastItemProgress = progressMap[playlist.last_engaged_on];
    const index = items.findIndex(i => (i.railcontent_id ?? i.id) === playlist.last_engaged_on);
    if (lastItemProgress?.status === 'completed') {
      nextId = items[index + 1] ?? nextId;
    } else if (lastItemProgress?.status === 'started') {
      nextId = items[index] ?? nextId;
    }
  }

  return {
    id: playlist.id,
    progressType: 'playlist',
    header: 'playlist',
    body: {
      first_items_thumbnail_url: playlist.first_items_thumbnail_url,
      title: playlist.name,
      subtitle: `${playlist.duration_formated} • ${playlist.total_items} items • ${playlist.likes} likes • ${playlist.user.display_name}`
    },
    progressTimestamp: item.progressTimestamp,
    cta: {
      text: 'Continue',
      action: {
          brand: playlist.brand,
          id: playlist.id,
          itemId: nextItem.id,
          type: 'playlists',
        };
    }
```

adrian-musora (commented):
I'm unable to test the code in-situ as I have no context for executing it. I ran through the code logic extensively and feel relatively comfortable that it executes as expected

Some bigger thoughts:

The logic for this is incredibly complicated. >.<  Just reading the docs took me a significant amount of time. 

MCS isn't designed to be responsible for packaging or formating user displayed information. Code from `processContenItem`.`getFormattedType()` and below would make more sense to be in the FEW or MA, ditto for the playlist formatting of the subtitle.

This needs to be optimized for performance before going to beta.

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+    .slice(0, limit + 5);
+}
+
+async function buildHierarchyMap(items) {
+  const contentIdsWithChildren = items
+    .filter(item => item.raw.child_count > 0)
+    .map(item => item.raw.id ?? item.raw.railcontent_id);
+
+  const entries = await Promise.all(
+    contentIdsWithChildren.map(async id => [id, await fetchHierarchy(id)])
```
adrian-musora, line 526: We're outside of my depth, so below might not be strictly true.

`fetchHierarchy` includes assignments in the hierarchy. In `processContentItem` we count the number of lessons that is returned by this. Unless I missunderstand the number of lessons will then include the assignment counts.
sanity documents include a `child_count` parameter (generated when we publish), this doesn't include  great/grandchildren count. So we'll be in a situatino where the user is shown "200 of 12 Lessons Completed" 
 
Currently assignments are completed when the user completes a lesson (tested in the system test today), but if this changes we'll have even stranger numbers.

 Chris' comment:
```Assignments exist, but don't have complete option and (supposedly) will be phased out in future content```

File: src/services/content.js [hunk 9]
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+
+  return {
+    id: data.id,
+    progressType: 'content',
+    header: contentType,
+    body: {
+      progressPercent,
+      thumbnail: data.thumbnail,
+      title: data.title,
+      subtitle: !data.child_count || data.lesson_count === 1
```
adrian-musora, line 655: `lesson_count` is defined as `child_count ?? 0` for all cases (contentTypeConfig.js and sanity.js) that I found and doesn't include children lessons. I think the correct value here is to sum the hiearchy count like `hierarchyMap.length`

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+      secondIncomplete = childProgress
+        .filter(([id]) => id !== firstIncomplete)
+        .find(([_, s]) => s !== 'completed')?.[0] || null;
+    }
+
+    data.completed_children = completedCount;
+    data.first_incomplete_child = childContentMap[firstIncomplete] || null;
+    if (data.type === 'pack') {
+      data.second_incomplete_child = childContentMap[secondIncomplete] || null;
+    }
```
adrian-musora, line 636: I couldn't figure this part out. Why do packs need the 2nd child? If i understand correctly we actually need to dig into the 2nd tier of `hierarchyMap` to grab the incompleted lessons as the map isn't flattened.

src/services/content.js
```diff
@@ -371,4 +373,320 @@ export async function getRecommendedForYou(brand, rowId = null, {
...
+    const childIds = Object.keys(children);
+
+    const childProgress = childIds.map(id => [id, progressOnChildren[id]])
+      .filter(([_, s]) => s != null).slice(1);
+
+    completedCount = childIds
+      .filter(id => childContentMap[id]?.child_count == null && progressOnLessons[id] === 'completed')
+      .length;
+
+    firstIncomplete = childProgress.find(([_, s]) => s !== 'completed')?.[0] || childProgress[0]?.[0];
```
adrian-musora, line 625: This needs different behaviour between guided courses and courses. 
Guided courses should always return the first incomplete lesson.
Courses should return the next lesson from the previous interacted regardless if there are incomplete lessons in the hierarchy.

Sourced from [here](https://www.figma.com/board/dDsiHGlAUNMVRZym2eEWyv/Progress-Logic?node-id=0-1&p=f&t=fV1AdvzjBNvB0wsA-0)

src/services/content.js
adrian-musora: Also part of a bigger optimization question

src/services/content.js
adrian-musora: also optimazation

PR #286, opened 2025-06-11
fix(BR-285): create cache service

alesevero:
## Summary by CodeRabbit

- **New Features**
  - Introduced an asynchronous cache service for improved data handling and consistency across the application.

- **Refactor**
  - Updated all cache operations to be asynchronous, ensuring consistent behavior and reliability.
  - Enhanced validation to ensure required configuration properties are present before initializing services.

- **Bug Fixes**
  - Improved timing and order of cache and timestamp updates, reducing potential data staleness issues.

### DIFF
--- a/src/services/cacheService.js
+++ b/src/services/cacheService.js
@@ -1,0 +1,15 @@
+import { globalConfig } from './config'
+
+const getItem = async (key) => {
+  return globalConfig.isMA ? await cache.getItem(key) : cache.getItem(key)
+}
+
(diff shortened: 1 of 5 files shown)


src/services/cacheService.js
```diff
@@ -0,0 +1,15 @@
+import { globalConfig } from './config'
```
robwillems, line 1: I'd call it localCacheService so its clearer across different use cases

PR #295, opened 2025-06-16
feat (beh 547) add course data for fe

adrian-musora:


src/services/sanity.js
```diff
@@ -1413,23 +1413,27 @@ async function fetchRelatedByLicense(railcontentId, brand, onlyUseSongTypes, cou
...
   const filterSongSameArtist = await new FilterBuilder(
-    `_type=="song" && _type==^._type && brand == "${brand}" && references(^.artist->_id) && railcontent_id !=${railContentId}`
+    `_type=="song" && _type==^._type && brand == "${brand}" && references(^.artist->_id) && railcontent_id !=${railContentId}`, {pullFutureContent: true}
   ).buildFilter()
   const filterSongSameGenre = await new FilterBuilder(
-    `_type=="song" && _type==^._type && brand == "${brand}" && references(^.genre[]->_id) && railcontent_id !=${railContentId}`
+    `_type=="song" && _type==^._type && brand == "${brand}" && references(^.genre[]->_id) && railcontent_id !=${railContentId}`, {pullFutureContent: true}
   ).buildFilter()
-  const filterNeighbouringSiblings = await new FilterBuilder(`references(^._id)`).buildFilter()
+  const filterNeighbouringSiblings = await new FilterBuilder(`references(^._id)`, {pullFutureContent: true}).buildFilter()
```
adrian-musora, line 1427: I think this is the only one, you'll want to check with FE/MA where this method is used, but based on the name of the variable, I believe this is the one we want to update

src/services/sanity.js
```diff
@@ -1413,23 +1413,27 @@ async function fetchRelatedByLicense(railcontentId, brand, onlyUseSongTypes, cou
...
   const childrenFilter = await new FilterBuilder(``, { isChildrenFilter: true }).buildFilter()
   const queryFields = `_id, "id":railcontent_id, published_on, "instructor": instructor[0]->name, title, "thumbnail":thumbnail.asset->url, length_in_seconds, web_url_path, "type": _type, difficulty, difficulty_string, railcontent_id, artist->,"permission_id": permission[]->railcontent_id,_type, "genre": genre[]->name`
   const queryFieldsWithSort = queryFields + ', sort'
   const query = `*[railcontent_id == ${railContentId} && brand == "${brand}" && (!defined(permission) || references(*[_type=='permission']._id))]{
-   _type, parent_type, railcontent_id,
+   _type, parent_type, 'parent_id': parent_content_data[0].id, railcontent_id,
+   'for-calculations': *[references(^._id)][0]{
+    'lessons-list': child[],
+    'courses-list': *[references(^._id)][0].child[]
```
adrian-musora, line 1435: From my understanding  will be called on the course-part or pack-bundle-lesson content type. So we need the count of siblings and parents, no the count of siblings and children.

PR #313, opened 2025-06-25
feat(TP 947): Add DataContext clearing on sign in/out

sheldonbaker:


src/lib/lastUpdated.js
```diff
@@ -29,6 +29,19 @@ export function wasLastUpdateOlderThanXSeconds(seconds, key) {
  *
  * @returns {void}
  */
-export function setLastUpdatedTime(key) {
-  globalConfig.localStorage.setItem(key, new Date().getTime()?.toString())
+ * @returns {Promise<void>}
+ */
```
alesevero, line 33: duplicated trailing comment block?

alesevero (approved):
Haven't tested it, but it looks good!

PR #325, opened 2025-07-04
Feat gc enroll

peterminhlee:


adrian-musora (approved):

src/services/content-org/guided-courses.ts
```diff
@@ -10,9 +10,9 @@ const excludeFromGeneratedIndex: string[] = []
 
 const BASE_PATH: string = `/api/content-org`
 
-export async function enrollUserInGuidedCourse(guidedCourse) {
+export async function enrollUserInGuidedCourse(guidedCourse, payload) {
```
adrian-musora, line 13: Not sure if this suggestion or a generic payload is better. Is there a prefered design?
```suggestion
export async function enrollUserInGuidedCourse(guidedCourse, {notifications_enabled = false}) {
```

PR #331, opened 2025-07-08
fix(sign-in): add account status and setup email functions

alesevero:


src/services/user/account.ts
```diff
@@ -0,0 +1,21 @@
...
+  const response = await httpClient.get<{ requires_setup: boolean }>(
+    `/api/user-management-system/v1/accounts/${email}/status`
+  )
+  return response
+}
+
+export async function sendAccountSetupEmail(email: string): Promise<void> {
+  const httpClient = new HttpClient(globalConfig.baseUrl)
+  return httpClient.post<void>(
+    `/api/user-management-system/v1/accounts/${email}/send-setup-email`,
```
sheldonbaker, line 18: `encodeURIComponent` could still be useful here - otherwise plus signs in an email could be misinterpreted (Laravel does it right, but some other layers in the future might not), 

```suggestion
    `/api/user-management-system/v1/accounts/${encodeURIComponent(email)}/send-setup-email`,
```

PR #333, opened 2025-07-09
feat(TP-947): v1 DataContext clearing

sheldonbaker:
## Summary by CodeRabbit

* **New Features**
  * Introduced new session management handlers for login and logout events.
  * Added the ability to clear all data contexts and user permissions data.
  * Exposed new functions for session and data context management in the public API.

* **Improvements**
  * Standardized all local storage and permissions operations to be asynchronous for better reliability.
  * Enhanced error handling and cache management for data contexts.

* **Tests**
  * Updated tests to properly handle asynchronous operations, ensuring accurate and reliable test results.


src/services/local-cache/index.js
```diff
@@ -0,0 +1,39 @@
+import { globalConfig } from '../../services/config.js'
```
robwillems, line 1: Suggestion:
On the frontend we'd like to minimize the amount of code where we can since its pulled to the client to run.  I think having a single LocalCache class with all the isMA logic in it is good enough.

  async getItem(key) {
  if(isMA) return await storage.getItem(key);
  return storage.getItem(key);
  }

PR #377, opened 2025-07-23
fix(BEH-873): Lessons Featuring this Song and Other Versions of this Song, restricted to current brand

adrian-musora:
[JIRA](https://musora.atlassian.net/browse/BEH-873)
ANd [this](https://musora.atlassian.net/browse/T3PS-377) which was tested but worked without me doing anything

## Description
- added missing brand restriction argument
- also updated yarn lock for building locally

## Testing

Published/Status restriction
I actually didn't make any changes for this. I checked the queryin the console log and saw the permissions + status restrictions. However, the results are cached briefly, so the bug is related to sanity caching, not the query. The sanity cache lasts a few minutes so you can unlist a piece of content, run the next test then come back :)

I used this as the base: drumeo/songs/transcription/260343
and this content which is related by license (which will show up if in published state and not if unlisted)
https://devapi.musora.com:9443/admin/studio/publishing/structure/__edit__learning-path-lesson_333874%2Ctype%3Dlearning-path-lesson
"C Color Chords"
ID: 333874

Brand restriction
On [Production](https://app.musora.com/drumeo/songs/transcription/260371) (has full version + guitareo versions) for both 
image

On [Local](http://localhost:5173/drumeo/songs/transcription/260371) (should only have the full version)
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filtering of related content to ensure results now match the specified brand as well as the content type.

### DIFF
--- a/src/services/sanity.js
+++ b/src/services/sanity.js
@@ -1435,9 +1435,8 @@ export async function fetchLessonsFeaturingThisContent(railcontentId, brand, cou
  */
 async function fetchRelatedByLicense(railcontentId, brand, onlyUseSongTypes, count) {
   const typeCheck = `@->_type in [${arrayJoinWithQuotes(SONG_TYPES)}]`
-  let typeCheckString = `brand == '${brand}' && `
+  let typeCheckString = `@->brand == '${brand}' && `
   typeCheckString += onlyUseSongTypes ? `${typeCheck}` : `!(${typeCheck})`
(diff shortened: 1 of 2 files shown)


eprachnau (commented):
Looks good! I also wasn't able to replicate T3PS-377

eprachnau (approved):

PR #383, opened 2025-07-26
fix(T3PS-139): fix lesson count on content cards

Meowbox64:
## Jira
- [T3PS-139](https://musora.atlassian.net/browse/T3PS-139)

## Changes
- updates `lesson_count` field to query documents filter based on user's account status (non/admin).
- fixes for:
  - tab data, for lessons Collections page/tab (lesson count pill)
  - content Rows, for For You page/tab and homepage recommended rows (lesson count pill)
  - homepage progress card subtitle ("X of Y lessons complete")
- admittedly this is a quick fix to make content cards show accurate lesson count for all the use cases I could think of. I basically copied the same query field to the different implementations, which feels wrong, but then again sanity.js is kinda messy.

## Testing
** for each of these cases please test with both an admin account (can see draft content) and plus account (non-admin, but has access to courses)
- use staging sanity dataset
- admin user - just use your account
- plus user - run `r sync-users --user-ids=843603 --source=beta-testing` to copy over the july8test2@musora.com user from beta-testing (same password as those)
- go to Lessons -> For You tab, under Perfect Your Pocket row, see Groove Coordination course. it should say 7 lessons if admin and 6 lessons if not.
- go to Collections tab, find Groove Coordination, and observe the same count(s)
- make progress one of its lessons
- go to homepage, find the course in progress row. It has a subtitle like "X of Y lessons complete". if admin it should say "X of 7 lessons complete" and if not then "X of 6 lessons complete"

[T3PS-139]: https://musora.atlassian.net/browse/T3PS-139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Lesson counts are now displayed more accurately across various content views.

* **Chores**
  * Removed unnecessary debug messages for a cleaner user experience.


src/services/sanity.js
```diff
@@ -565,6 +566,7 @@ export async function fetchContentRows(brand, pageName, contentRowSlug)
   if (pageName === 'lessons') pageName = 'lesson'
   if (pageName === 'songs') pageName = 'song'
   const rowString = contentRowSlug ? ` && slug.current == "${contentRowSlug.toLowerCase()}"` : ''
+  const lessonCountFilter = await new FilterBuilder(`_id in ^.child[]._ref`, {pullFutureContent: true}).buildFilter()
```
alesevero, line 569: If you foresee this and the same code above to be in sync I'd extract it so we avoid changing one but not the other. Otherwise we can leave it like this

adrian-musora (approved):

PR #391, opened 2025-07-31
fix(MU2-907): scheduled content appearing in new releases

slumrig:
Content with published_on date in the future were appearing in new releases. Removed some redundant filters and added check that published_on is on or before current date.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the accuracy of new releases by updating filtering logic to display only published content available as of today.
  
* **Chores**
  * Updated internal logic for fetching new releases to enhance reliability and consistency.


src/services/sanity.js
```diff
@@ -350,8 +350,8 @@ export async function fetchNewReleases(
   const start = (page - 1) * limit
   const end = start + limit
   const sortOrder = getSortOrder(sort, brand)
-  const nextQuarter = getNextAndPreviousQuarterDates()['next']
-  const filter = `_type in ${typesString} && brand == '${brand}' && status == 'published' && show_in_new_feed == true && (!defined(quarter_published) ||  quarter_published != '${nextQuarter}')`
+  const now = new Date().toISOString().split('T')[0]
```
eprachnau, line 353: This might be a good candidate for a helper function in the future if we end up filtering by date more.
`getNextAndPreviousQuarterDates` is dead code now, we could delete it

eprachnau (approved):

PR #409, opened 2025-08-07
fix(BEH-601): fix the over encompassing previous change

ereztdev:
## Jira
- no ticket

## Changes
- fix unintentional sanity query prefix overwrite, affecting all sanity queries that have child filter across brand site
- also fix fetchSiblingContent query incorrectly referencing enclosing document scope

## Testing
- link mcs
- go to a pack bundle lesson and verify you get a collection module that has sibling lessons in that pack-bundle, with correct data about index (like "Course X of Y, Lesson A of B")
- load homepage and content cards should load (previously the fetch queries returned null)

## Summary by CodeRabbit

* **Refactor**
  * Improved the logic for assigning filter prefixes, ensuring more consistent and accurate filter behavior.
  * Updated how parent lists are filtered when fetching sibling content, potentially affecting the display of future or parent-related items.

### DIFF
--- a/src/filterBuilder.js
+++ b/src/filterBuilder.js
@@ -36,8 +36,17 @@ export class FilterBuilder {
     this.filter = filter
     // this.debug = process.env.DEBUG === 'true' || false;
     this.debug = false
-    this.prefix = isChildrenFilter ? '@->' : ''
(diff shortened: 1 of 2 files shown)


Meowbox64 (approved):
good

PR #243, opened 2025-04-29
feat(BEH-402) Playlist V2 - Endpoints for like/unlike/duplicate/report/reorder

adrian-musora:
[MPB PR](https://github.com/railroadmedia/musora-platform-backend/pull/68)

Pin/unpin endpoints not exposed as they are not currently part of the design. 
Update endpoint has been added in this [PR](https://github.com/railroadmedia/musora-platform-backend/pull/67), but work is still ongoing

## Summary by CodeRabbit

- **New Features**
	- Introduced new playlist management options, including duplicating, liking, unliking, reporting, reordering, and toggling playlist privacy.
	- Added ability to check if content is liked by IDs and to fetch like counts.
	- Enabled querying of resume time for content by IDs.
- **Improvements**
	- Simplified playlist fetching by removing certain filters.
	- Updated documentation for playlist creation.
- **Removals**
	- Removed support for a deprecated content context feature.
	- Consolidated playlist-related functions by removing redundant exports and implementations from other modules.

PR #411, opened 2025-08-07
fix(MU-922): uses BE for fetching recent content items

sheldonbaker:
[see mpb PR](https://github.com/railroadmedia/musora-platform-backend/pull/243)

## Summary by CodeRabbit

- New Features
  - Added tab-scoped “Recent” browsing (All, Completed, Incomplete) per page.

- Refactor
  - Lesson content now loads faster via parallel requests.
  - More consistent pagination and recent-results behavior across tabs.
  - Reduced payload for tab content to improve responsiveness.

- Tests
  - Updated tests to match the redesigned recent API and expected results.


src/services/content.js
```diff
@@ -5,22 +5,21 @@
...
 } from './sanity.js'
+import {getRecentTypesForPage} from '../contentTypeConfig.js'
 import {TabResponseType, Tabs, capitalizeFirstLetter} from '../contentMetaData.js'
-import {fetchHandler} from "./railcontent";
+import {fetchRecent} from "./railcontent";
 import {recommendations, rankCategories, rankItems} from "./recommendations";
 import {addContextToContent} from "./contentAggregator.js";
 
-
 export async function getLessonContentRows (brand='drumeo', pageName = 'lessons') {
```
adrian-musora, line 21: These should not be defaulted and should always be populated by the caller
```suggestion
export async function getLessonContentRows (brand, pageName) {
```

adrian-musora (approved):

src/services/content.js
```diff
@@ -159,13 +151,39 @@ export async function getTabResults(brand, pageName, tabName, {
  *   .then(content => console.log(content))
  *   .catch(error => console.error(error));
  */
-export async function getRecent(brand, pageName, tabName = 'all', {
+export async function getRecent(brand, { status, types } = {}, {
```
adrian-musora, line 154: The method prototype has changed and the contentProgress.test.js tests will now fail. I don't have them running on my computer.... which is a problem on my end. But they will crash immediately now.

PR #443, opened 2025-09-09
feat(MU2-998): add customer order endpoint

slumrig:
New function for endpoint for customer order payments.

[MU2-988](https://musora.atlassian.net/browse/MU2-998)
[MPB PR](https://github.com/railroadmedia/musora-platform-backend/pull/279)

## Summary by CodeRabbit

* **New Features**
  * Added a public API to retrieve customer payment records (initial implementation; no pagination or advanced error handling yet).

* **Chores**
  * Updated repository ignore settings to exclude the lockfile from version control.


src/services/payments.js
```diff
@@ -0,0 +1,15 @@
...
+
+export async function fetchCustomerPayments(brand, { page = 1, limit = 10 } = {}) {
+  const params = {
+    method: 'GET',
+    headers: {
+        'Authorization': `Bearer ${globalConfig.sessionConfig.authToken}`,
+        'Content-Type': 'application/json',
+      }
+    }
+    return await fetch(globalConfig.baseUrl + '/api/customer/orders/v1', params)
```
roxanariza, line 14: You can use fetchHandler method here

src/services/payments.js
```diff
@@ -0,0 +1,10 @@
+/**
```
alesevero, line 1: we should be implementing it with typescript. Don't forget to add types and the jsdoc (which should still be used for now).

After docs are added, please run `npm run doc` to update the docs for MCS with this new function

src/services/payments.js
```diff
@@ -0,0 +1,10 @@
...
+ * @module Payments
```
alesevero, line 2: I feel this should be in a subfolder alongside other membership/ecommerce functionality. We can iterate on it to decide on what better place would be.

src/services/payments.js
```diff
@@ -0,0 +1,10 @@
...
+ */
+import { globalConfig } from './config.js'
+import { fetchHandler } from '../lib/httpHelper.js'
+
+export async function fetchCustomerPayments() {
+    const response = await fetchHandler(globalConfig.baseUrl + '/api/customer/orders/v1', globalConfig.sessionConfig.authToken, 'get')
```
alesevero, line 8: I recommend using HttpClient when implementing it with typescript, unless you need access to http headers from the response.

alesevero (approved):

PR #455, opened 2025-09-18
feat(MU2-1038): New method for follow/unfollow threads

roxanariza:
New methods for follow/unfollow threads


src/services/forums/threads.ts
```diff
@@ -28,3 +28,27 @@ export async function createThread(
...
 }
+
+/**
+ * Follow a thread.
+ *
+ * @param {number} threadId - The ID of the thread to lock.
+ * @return {Promise<void>} - A promise that resolves when the thread is locked.
+ * @throws {HttpError} - If the request fails.
+ */
+export async function followThread(threadId: number): Promise<void> {
```
csillaj, line 39: [req] you forgot to export these methods in index

csillaj (changes requested):

PR #596, opened 2025-11-27
chore(AGI): update to use typescript sanity client

alesevero:
_Note: this PR depends on #646_

## Description/Design
Using BigBoggy's ™️  Sanity typescript client to improve type-safety and composability for sanity data querying and manipulation.

## Changes 
- Update Artists, Genre and Instructors to make usage of our typescript sanity client
- Add `DocumentType` union
  - based on `array::unique(*[]._type)` minus the following:
```
  's3-files.storedFile',
  'sanity.assist.schemaType.annotations',
  'sanity.fileAsset',
  'sanity.imageAsset',
  'system.group',
  'system.release',
  'system.retention',
```
- Change `fetchSingle` to `fetchFirst`
- Add new `fetchSingle` to `SanityClient.ts`
- Make `SanityClient.fetchList` return standardized data with pagination data to the response
```typescript
export interface SanityListResponse<T = any> {
  data: T[]
  total: number
  sort?: string
  offset?: number
  limit?: number
}
```
- Simple response decorators at the expense of a bit of loose types for now:
```typescript
return contentClient
  .fetchList<Lesson>(q, {
    sort,
    offset,
    limit,
  })
  .then(needsAccessDecorator)
```

## Next Steps (in future PRs)
- (breaking change) Add errors as values from #349: #600  _(probably not)_
- Decorator builder for parallel and sequencial decorator processing


src/infrastructure/sanity/clients/ContentClient.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    // Use custom fields if provided, otherwise use defaults
+    let fields = customFields || defaultFields
+
+    // Add children-related fields if requested
+    if (includeChildren) {
+      fields = [
+        ...fields,
+        "'child_count': coalesce(count(child[]->), 0)",
+        `"lessons": child[]->{
```
Meowbox64: my opinion is that this should be named "children", but as long as we have a uniform child field naming goal (child,children,lessons,etc) I'm happy

Meowbox64 (approved):
lgtm

src/infrastructure/sanity/clients/ContentClient.ts
```diff
@@ -0,0 +1,174 @@
...
+      const q = query()
+        .and(brand ? `brand == "${brand}"` : '')
+        .and(`_type == "${type}"`)
+        .order(sortBy)
+        .slice(offset, offset + limit)
+        .select(...(fieldsString ? [fieldsString] : []))
+        .build()
+
+      return this.fetchList<T>(q, {
+        sort: sortBy,
```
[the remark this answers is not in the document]
alesevero: That's because fetchList is the actual query executor, and it adds that info (sort, limit, offset) into the response object.

Since fetchByTypeAndBrand is just a helper, it needs to pass those props down to the main function.

src/infrastructure/sanity/clients/ContentClient.ts
alesevero: I agree. I'll will remove this ContentClient file.

With the new query and filter builders, this class kinda loose its purpose (abstract queries) + adds a bunch of hidden complexity (like this one you're commenting.)

PR #772, opened 2026-02-04
refactor: refactor part of sanity.js

Meowbox64:
## ticket
- sanity.js is huge and needs to be broken up and refactored
  - like with the AGI functions :chef_kiss:
- found a common group of functions: scheduled
  - in charge of fetching newly released and scheduled-to-be-released content and events
  - functions used on homepage "New and Upcoming" content row, and the content rows on the scheduled page
- also found many helper functions in sanity.js, which are also perfect to go into their own file -> helper.ts
- also also decided that with the move to split up sanity.js, we should also separate out `fetchSanity` and corresponding functions, leaving sanity.js with only fetch functions, and is now just another fetch function file like the AGI files

## Design
- `scheduled.ts` follows the design of the Artist,Genre,Instructor (AGI) ts files, which used to also live in sanity.js
  - common fields moved to a const :ok_hand: 
  - common filters moved to a filter generator fn (could argue it's complexity isn't much better than hard-coded queries in each fetch fn)
- common sanity.js helper functions moved to `helper.ts` (sanity.js now imports those fn's from this file)
- moved `fetchSanity` and related internal fn's to `fetch.ts`, and updated all imports (minus tests, because those need their own refactor)

## Platform feature changes
- homepage "New and Upcoming" row now:
  - returns currently-airing live events
  - filters new and upcoming (not live) contents by `show_in_new_feed == true`

## Testing
- use staging dataset

quick:
- go to homepage -> New and Upcoming content row
- see "PlayBass Launch Q&A With Chris Wong" (only one that matches the new logic)
- go to https://devapp.musora.com:5174/drumeo/schedule
- take a screenshot of the contents
- revert mcs changes, and see that the cards are all the same

thorough:

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -38,6 +38,13 @@ import {
 } from './services/content/instructor.ts';
 
 import {
+	fetchNewReleases,
+	fetchNewUpcomingAndLive,
+	fetchScheduledReleases,
+	fetchUpcomingEvents
+} from './services/content/scheduled.ts';
+
+import {
 	enrollUserInGuidedCourse,
 	fetchEnrollmentPageMetadata,
 	guidedCourses,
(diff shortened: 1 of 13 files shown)


Meowbox64:
too old to be useful, and will be tackled with the rest of mcs refactors

PR #867, opened 2026-03-17
fix: remove flush session interval

sheldonbaker:
[Related MPF PR](https://github.com/railroadmedia/musora-platform-frontend/pull/1534), which no longer needs the push interval to be handled in MCS.

(MA doesn't use the intervals).


slumrig (approved):
yes 👉

PR #889, opened 2026-03-27
feat(BEHSTP-167): offline progress tracking support

Meowbox64:

- https://musora.atlassian.net/browse/BEHSTP-167

## Problem

- MA needs ability to read/write practices, content progress, and user activity (PCPA) offline. watermelon solves this, but some functions have a fetch sanity baked in in order to enrich data. This PR refactors those functions and creates "_Offline" versions which return un-enriched watermelon data.
- MA download endpoints `fetchLessonContent` and `fetchByRailcontentIds` need to return hierarchy data, so content metadata is accessible offline for progress writes.

## Description/Design

progress should not bubble offline
- refactors PCPA functions. creates `_recordWatchSession` function that contain the actual logic, and `recordWatchSession` and `recordWatchSessionOffline` call the respective call the internal one differently. this design is used across several refactored functions.
- MA uses `fetchLessonContent` for sinlge content downloads: modifies `fetchLessonContent` with a `forDownload` param, which does a postprocess to add hierarchy to object response, allowing MA to download it along with content without changing downloads behaviour.
- MA uses `fetchByRailcontentIds` for multiple content downloads: calls getHierarchies if type=downloads to hydrate response contents with their respective hierarchies
- refactors and unifies downloads response fields between the 2 functions
- offline user activity functions is really just getting RADFOP: Recent Activities Derived From Offline Progress.
- fixed regex parsing for `npm run build-index` excluding functions

## Testing

use devendpoint to test.

### Smoketest existing online functions
- homepage progress row loads.
- profile page displays accurate calendar, streak stats, practice circle, practice list, and user activities.
some of these read functions have been slightly refactored, so its good to just quickly check they still work.

### Single Content Metadata Extraction + recordWatchSessionOffline
```
fetchLessonContent(415432, {forDownload: true}).then((r) => {
    const hierarchy = r.hierarchy
    console.log('metadata', hierarchy.metadata[415432])
    recordWatchSessionOffline(415432, 300, 60, 60, hierarchy).then((status) => {
      console.log('status', status)
      getProgressDataByIds([415432, 415525]).then((progress) => {
        console.log('progress', progress)
      })
    })
  })
```
- verify fetchLessonContent:
  - verify metadata is correct
- verify contentStatusCompleted
  - verify progress write response for 415432 has content_brand, content_parent_id, content_type hydrated
  - verify getProgressDataByIds response has progress for ONLY the child (415432), and parent 415525 is still 0% progress (not bubbled)

### Multi-Content Metadata Extraction
```
fetchByRailContentIds([420219, 415431], 'download').then((r) => {
  .forEach((c) => {
    console.log(`metadata ${c.id}`, c.hierarchy.metadata[c.id])
  })
})
```
- verify `metadata` is correct for both contents

### getRecentActivityOffline
```
const offlineTimestamp = Date.now() - 3600 * 1000
  contentStatusStarted(415432).then((s1) => {
    getRecentActivityOffline(offlineTimestamp).then((started) => {
      console.log('RADFOP after started action:', started)
      contentStatusStarted(415432).then((s2) => {
        getRecentActivityOffline(offlineTimestamp).then((completed) => {
          console.log('RADFOP after completed action:', completed)
        })
      })
    })
  })
```
- 'RADFOP after started action:' should show start action activity
- 'RADFOP after completed action:' should show start and complete action activities
- note that if you reset progress on a content, the corresponding ~user activity~ RADFOP will disappear, and thats ok for now.

### getPracticeSessionsOffline
```
const offlineTimestamp = Date.now() - 3600 * 1000
  trackUserPractice(415432, 10).then((status) => {
    getPracticeSessionsOffline(offlineTimestamp).then((data) => {
      console.log('Recent activity offline:', data)
    })
  })
```
- returns correct practices data without corresponding metadata

## notes:

- not sure how to properly test offline capabilities on web (dont really want to go through the hassle of setting up MA test for this), but you could get `fetchLessonContent().hierarchy` while online, then pass it in to contentStatusCompleted while offline and verify it works. I'm confident it does.


sheldonbaker (approved):

adrian-musora (approved):

src/services/offline/activities.ts
```diff
@@ -0,0 +1,174 @@
...
+      contentType: type,
+      date: dayjs(p.created_at).format('YYYY-MM-DD'),
+      brand: p.content_brand,
+    })
+    if (p.state === STATE.COMPLETED) {
+      activities.push({
+        contentId: p.content_id,
+        action: 'complete',
+        contentType: type,
+        date: dayjs(p.updated_at).format('YYYY-MM-DD'),
```
sheldonbaker: Not sure if there's anything else we can do (aside from adding and setting `completed_at` timestamps), but this is fragile - the `updated_at` could (theoretically) be changed well after the content was completed.

src/services/contentProgress.js
```diff
@@ -469,41 +550,54 @@ async function trackProgress(contentId, collection, currentSeconds, mediaLengthS
     // doesn't affect livestream resumeTime, but will send users to 0 seconds in VOD
     currentSeconds = 0
   }
-  return saveContentProgress(contentId, collection, progress, currentSeconds, { skipPush: true })
+  return saveContentProgress(contentId, collection, progress, currentSeconds, { hierarchy, skipPush: true })
```
sheldonbaker: Does passing `hierarchy` here make `saveContentProgress` think it's in offline mode?

Maybe consider a more explicit `offline` flag in those cases (instead of checking `if (hierarchy)`?

src/services/offline/activities.ts
```diff
@@ -0,0 +1,174 @@
+import { db } from '../sync'
```
alesevero, line 1: **[nit]:** already have an `offline` directory where this file is on, so I'd split it based on functionality/domain (activity.ts, practices.ts, etc.)

src/services/sanity.js
```diff
@@ -536,6 +536,18 @@ export async function fetchByRailContentIds(
       }
       return result
     }
+    const chapterProcess = (result) => {
+      const chapters = result.chapters ?? []
+      if (chapters.length === 0) return result
```
MeenuDhanoa: We also need to map chapters on the children. Updating it like this should fix the issue.

```
      if (chapters.length > 0) {
        result.chapters = chapters.map((chapter, index) => ({
          ...chapter,
          chapter_thumbnail_url: `https://musora-web-platform.s3.amazonaws.com/chapters/${result.brand}/Chapter${index + 1}.jpg`,
        }))
      }
      const children = result.children
      if (Array.isArray(children) && children.length > 0) {
        result.children = children.map((child) => chapterProcess(child))
      }
```

PR #903, opened 2026-04-03
feat(BEHSTP-169): end screen method

roxanariza:

- [BEHSTP-169](https://musora.atlassian.net/browse/BEHSTP-169)
## Description/Design

Implement end screen logic using new MCS method getEndScreen.

Covers all scenarios for collections, course collections, single lessons, and playlists:

- Handles “Up next” behavior (next lesson vs recommendation).
- Differentiates between countdown and course complete variants.
- Integrates RecSys highest-confidence recommendations.
- Adds fallback logic when no recommendations are available.

Ensures consistent end screen behavior across all content types.
Integrates learning path end screen logic via new MCS getLearningPathEndScreen  
Migrates endScreen service to TypeScript with separated types and [README ](https://github.com/railroadmedia/musora-content-services/blob/BEHSTP-169-end-screen-method/src/services/endScreen/README.md)

## Testing

- can run tests:  npx jest test/endScreen.test.js --no-coverage

## Additional Notes

- _Provide any additional context or notes that might be helpful for the reviewer._

[BEHSTP-169]: https://musora.atlassian.net/browse/BEHSTP-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/endScreen/endScreen.ts
```diff
@@ -0,0 +1,222 @@
...
+// ─── Public API ───────────────────────────────────────────────────────────────
+
+export async function getEndScreen({
+  lesson,
+  course = null,
+  collection = null,
+  playlist = null,
+  brand,
+  learningPath = null,
+  lessonWasPreviouslyCompleted = false,
```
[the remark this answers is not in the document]
roxanariza: I looked into calculating it internally from learningPath, but there's a timing issue that makes it tricky. If  getLearningPathEndScreen needs to be called after the lesson is marked complete (reactively on FE, explicitly on MA after progress update), the lesson already shows as completed in the data, so there's no way to distinguish "completed in this session" from "already completed before" without the initial snapshot.  That's why it needs to come from outside, when you fetch lesson data.

src/services/endScreen/README.md
```diff
@@ -0,0 +1,173 @@
...
+    previous_learning_path_dailies?:    { id: number, progressStatus?: string }[]
+    next_learning_path_dailies?:        { id: number, progressStatus?: string }[]
+  }
+  lessonWasPreviouslyCompleted?: boolean  // used with learningPath, default false
+}
+
+// Return value
+{
+  variant:           'countdown-up-next' | 'course-complete' | 'path-complete' | 'what-to-do-today'
+  upNext:            object | null  // null only for path-complete and method-session-complete
```
Meowbox64: is `method-session-complete` a state? It looks like its only referenced here in this README. And this doesnt make sense; if the daily session is complete, the user gets directed to the next incomplete lesson of the LP

src/services/endScreen/types.ts
```diff
@@ -0,0 +1,82 @@
+/**
+ * @module EndScreen Types
+ */
+
+export type EndScreenVariant =
+  | 'countdown-up-next'
+  | 'course-complete'
+  | 'path-complete'
```
Meowbox64, line 8: is this also used for completing a guided course that has an award? if not, which state? or do GC's simply give you award via toast, and the endscreen behaviour is the same as normal courses?

src/services/endScreen/endScreen.ts
```diff
@@ -0,0 +1,216 @@
...
+async function fetchEndScreenRecommendation(
+  brand: string,
+  contentId: number,
+  excludeId: number | null = null
+): Promise<any | null> {
+  try {
+    let ids: number[] = await fetchSimilarItems(contentId, brand, 50)
+    if (!Array.isArray(ids) || ids.length === 0) {
+      //TODO: will be defined in config, need to decide with Chris about the ids
+      ids = [373201]
```
Meowbox64, line 200: make a function `getFallbackId`?

src/services/endScreen/endScreen.ts
```diff
@@ -0,0 +1,216 @@
...
+      contents?.find((c) => c.id !== excludeId && (!c.parent_id || c.parent_id !== excludeId)) ??
+      null
+    if (!recommended) return null
+    return await addContextToContent(() => recommended, {
+      addProgressPercentage: true,
+      addProgressStatus: true,
+      addNavigateTo: true,
+    })
+  } catch {
+    return null
```
Meowbox64, line 214: is this the best choice? it means that we'll be building valid response objects only where `upNext` is null.  FE/MA would have to look into the object determine failure, and they'd be left with an error state and therefore no endscreen.

I'd say if this returns null, then we need to handle it better at the caller. If errors bubble too far, can we catch and return an object with `EndScreenVariant`=`error` and return with an `ERROR_CTA` having simply a "go back" button, so to the user nothing seems broken? I'd love to chat more about this

src/services/endScreen/endScreen.ts
```diff
@@ -0,0 +1,216 @@
...
+  ...jamTrackLessonTypes,
+  ...lessonTypesMapping['single lessons'],
+]
+const COURSE_COMPLETE_CTA: CtaLabels = { primary: 'Play Now', secondary: 'Back to Home' }
+const COUNTDOWN_CTA: CtaLabels = { primary: 'Play Now', secondary: 'Cancel' }
+const PATH_COMPLETE_CTA: CtaLabels = { primary: 'View Achievement', secondary: 'Back to Home' }
+
+// ─── Public API ───────────────────────────────────────────────────────────────
+
+export async function getEndScreen({
```
Meowbox64, line 31: thought: how are endscreens supposed to function offline? Do they use the playlist flow? Ask Chris. We maybe also want to guard against attempted fetches if offline.

src/services/endScreen/endScreen.ts
```diff
@@ -0,0 +1,216 @@
...
+  const children = learningPath.children ?? []
+  const dailies = learningPath.learning_path_dailies ?? []
+  const previousDailies = learningPath.previous_learning_path_dailies ?? []
+  const nextDailies = learningPath.next_learning_path_dailies ?? []
+
+  const allLessonsCompleted =
+    children.length > 0 && children.every((child) => child.progressStatus === 'completed')
+
+  if (allLessonsCompleted && !lessonWasPreviouslyCompleted) {
+    return buildPathComplete()
```
Meowbox64, line 98: shouldnt we also be checking if the user has some daily session lessons in the next learning path? and if so, show them the next lesson and autoplay for them?

src/services/endScreen/endScreen.ts
```diff
@@ -0,0 +1,216 @@
...
+  if (learningPath.is_active_learning_path) {
+    // Exclude next LP lessons when checking if today's session is done
+    const currentDailies = effectiveDailies.filter((d) => !d.isNextLP)
+    const dailyLessonsCompleted =
+      currentDailies.length > 0 && currentDailies.every((d) => d.progressStatus === 'completed')
+
+    if (!dailyLessonsCompleted) {
+      const nextDailyLesson = getNextInArrayOrNull(lesson.id, effectiveDailies)
+      if (nextDailyLesson) {
+        return buildWhatToDoToday(nextDailyLesson)
```
Meowbox64, line 121: same comment as above

Meowbox64 (commented):
i think there's some learning paths logic errors here

src/services/endScreen/README.md
roxanariza: The method-session-complete reference in the README comment is leftover from the initial implementation. It was later updated to support the three end screen states:

Active LP + dailies not completed → What to do today
Active LP + all dailies completed / inactive LP → Up next
All lessons completed → Path complete

src/services/endScreen/endScreen.ts
roxanariza: Based on Peter's comment we should not check next learning path

Image

roxanariza: I will ask Chris, to be sure

PR #920, opened 2026-04-15
fix: loki adapter cjs/esm export issues in mcs-cli

sheldonbaker:
fix mcs-cli error when importing loki adapter in node


slumrig (approved):
yes 👉

PR #938, opened 2026-04-23
fix: remove timestamp filtering of practice records in offline mode

sheldonbaker:

[BEHSTP-232](https://musora.atlassian.net/browse/BEHSTP-232)

## Description/Design

Returns all practice records from watermelon in offline methods. Responsibility for correct display of said records (when no metadata like title, thumbnail, etc. available) falls to MA.

[BEHSTP-232]: https://musora.atlassian.net/browse/BEHSTP-232?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/services/offline/activities.ts
+++ b/src/services/offline/activities.ts
@@ -20,7 +20,7 @@ interface Activity {
  * @returns `{ currentPage, totalPages, data }` where `data` is a page of `Activity` records
  */
 export async function getRecentActivityOffline(
-  offlineTimestamp: number,
+  _offlineTimestamp: number,
   {
     page = 1,
     limit = 5,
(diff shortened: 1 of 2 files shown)


src/services/offline/activities.ts
```diff
@@ -54,9 +54,8 @@ export async function getRecentActivityOffline(
   }
 }
 
-function getClauses(offlineTimestamp: number, tabName: string|null) {
+function getClauses(tabName: string|null) {
   let clauses: Q.Clause[] = [
-    Q.where('updated_at', Q.gte(offlineTimestamp)),
     Q.sortBy('created_at', 'desc'),
```
Meowbox64, line 59: need a `date` check here?

Meowbox64 (approved):

PR #952, opened 2026-05-04
fix(BEHSTP-160): hide future content in related lessons

adrian-musora:

[JIRA](https://musora.atlassian.net/browse/BEHSTP-160?focusedCommentId=29451)

## Description/Design

- I only show published content (in the past) no more extra content
- In order to show future content (scheduled) given the default status, we don't restrict publish_on. By explicitly setting `published` we include the published_on restriction.

## Testing

- ` mcs call fetchRelatedLessons [411128]` and I logged the query to verify the 
- I trust the query :D. Testing this requires setting up more stuff in sanity that I can to do.


slumrig (approved):
yes 👉

PR #955, opened 2026-05-06
feat(BEHSTP-304): LP lesson downloads & refactors

Meowbox64:

- https://musora.atlassian.net/browse/BEHSTP-304

### Features:
- return `learning_path_parent_id` on MA-download-fetch functions (fetchLessonContent & fetchByRailcontentIds) for MA to be able to tell if there is a LP parent.
  - necessary for setting `collection` when saving progress offline

### Fixes:
- fixed type issues
- fixed missing awaits, causing progress duplication issues
- fixed skip push issues
- update `getExclusionList` for global index file to be able to read multi-line `excludeFromGeneratedIndex` arrays
- fix addContextToContent for LP collection lessons
  - collection parameter was previously incorrectly missing, meaning all progress reads read from a-la-carte lesson instances
- fixed `getHierarchies` caller -> extracts metadata correctly
- fixed progress bubbling to parent tying to set progress lower than existing (now works)

### Refactors:
- split up online/offline write function logic into separate functions, for clean logic's sake
  - offline counterparts moved to `offline/progress.ts`
  - affected `contentStatusReset`, `contentStatusStarted`, `contentStatusCompleted`
  - `saveContentProgress` is too deeply nested, and must remain with a `isOffline` flag
  - `handleLearningPathProgressActions` was refactored to have an offline flag for differing logic
- consolidated functions with duplicate logic
- moved validation logic for `onContentCompletedLearningPathActions` to its caller and renamed fn to `onLearningPathCompletedActions`
- add unit and integration tests covering ALL functions in contentProgress.js (its about time :green_circle:)
  - unfortunately, to effectively test fn's in isolation, each function must be exported. see note 1 at bottom.

### Chores: 
- removed unintentionally exported internal functions from export

### Automated tests!
you're probably fine to skip reviwing these if you want. I spent a whole day going through all of them to ensure they're correct.
- `npm test unit/content-progress`
- `npm test:integration integration/content-progress`
  - includes `e2eProgress.test.ts`, which are intended to check all side effects of progress write functions!

### Manual Testing:
- smoke test all progress-write areas:
  - setting progress to 50%, then completing, then resetting on an individual lesson/song
  - doing the above to a course lesson and verifying 
  - verify bubble/trickles work properly for course collection courses
  - verify playlist progress writes to a-la-carte record
  - complete an LP:
    - verify all progress writes bulk into 1 call (about 60 records)

## Additional Notes

`1 - ` in order to test individual functions in isolation, especially unit test pure logic on internal helper functions, those functions must be exported. Only problem is then the global index will read it and export to package public API. Bad. But also i don't want to only test internal pure logic function _through_ their non-pure callers. so I'm stuck here, and just went with the easy-but-clunky solution of manually excluding each function that shouldnt be exported from contentProgress :nerd_face:

`2 - ` Certain service class functions are great to have reusable mocks for. getSanityDate for example. So i moved them into a directory inside /tests/integration/content-progress/__mocks__/. other options:
- `/tests/__mocks__/` if these are safe to be reused across all module test mocking
- `/src/services/__mocks__/sanity.mock.ts` each file gets it's own mock file placed in the services directory, alongside the other real file.

`3 - ` to be fair, the MA changes can be separated from the refactor changes. but I came across several bugs throughout refactoring too, so I'd need to find and extract those. and then since I'm changing logic, I'd have to write tests...


src/services/offline/progress.ts
```diff
@@ -1,19 +1,29 @@
...
+  filterOutLearningPathsForDuplication,
+  filterOutNegativeProgress,
+  getProgressDataByIds,
+} from '../../services/contentProgress'
+import { COLLECTION_ID_SELF, COLLECTION_TYPE, CollectionParameter } from '../sync/models/ContentProgress'
+import { db } from '../../services/sync'
+
+const excludeFromGeneratedIndex = [
+  'duplicateProgressToALaCarteOffline',
+]
```
slumrig, line 14:
```suggestion
```
import not used

src/services/contentProgress.js
```diff
@@ -599,262 +641,229 @@ async function saveContentProgress(
...
+  if (collection?.type !== COLLECTION_TYPE.LEARNING_PATH) {
+    return
+  }
+
+  await duplicateProgressToALaCarte(progresses, collection)
+
+  for (const [id, prog] of Object.entries(progresses)) {
+    if (
+      prog === 100
+      && collection?.type === COLLECTION_TYPE.LEARNING_PATH
```
slumrig:
```suggestion
```

return early above on same check so is redundant here

test/integration/content-progress/e2eProgress.test.ts
```diff
@@ -0,0 +1,603 @@
...
+  return result.data
+}
+
+async function writeOne(contentId: number, progress: number, collection: CollectionParameter = null) {
+  await db.contentProgress.recordProgress(contentId, collection, progress, testMetadata, 0, {skipPush: true})
+}
+
+async function writeSome(contentIds: Record<string, number>, collection: CollectionParameter = null) {
+  const contentMetadataMap = Object.fromEntries(Object.keys(contentIds).map(id => [id, testMetadata]))
+  await db.contentProgress.recordProgressMany(contentIds, collection, contentMetadataMap, {skipPush: true})
```
slumrig: nit: `{ } ` sometimes you add spaces after and before the brackets, not sure what our standard is but keep it consistent - e.g. line 82
`{ id: 100, type: 'lesson' }`

test/integration/initializeTestDB.ts
```diff
@@ -0,0 +1,75 @@
...
+    initializeTestService()
+    const userScope = makeUserScope()
+    SyncTelemetry.setInstance(makeTelemetry(userScope))
+    const manager = new SyncManager(userScope, makeContext(), () => makeDatabase())
+    teardown = SyncManager.assignAndSetupInstance(manager)
+    ctx.pushSpies = initializeMockPushes()
+  })
+
+  afterEach(async () => {
+    for (let i = 0; i < 200; i++) await new Promise(resolve => setImmediate(resolve))
```
slumrig, line 72: not: why 200? magic number?

src/services/sanity.js
```diff
@@ -976,7 +976,7 @@ export async function fetchLessonContent(railContentId, { forDownload = false }
     showMembershipRestrictedContent: true,
   }
 
-  const fields = getFieldsForContentType('download')
+  const fields = getFieldsForContentType('download') // a later refactor can make this 'playback'.
```
slumrig: nit: TODO with a ticket number for future refactor (I realize I'm bad at this too :) )

slumrig (approved):
Some nitpicks but nice refactoring and test coverage adds to the confidence level for this!

src/services/offline/progress.ts
Meowbox64: need this for blocking `duplicateProgressToALaCarteOffline` from being included in the global index export file

test/integration/initializeTestDB.ts
Meowbox64: buffer number, gives enough time for promise chains to settle

src/contentTypeConfig.js
```diff
@@ -48,24 +48,24 @@ export const artistField = `select(
...
-  "'length_in_seconds' : coalesce(length_in_seconds, soundslice[0].soundslice_length_in_second)",
+  '\'type\': _type',
+  '\'length_in_seconds\' : coalesce(length_in_seconds, soundslice[0].soundslice_length_in_second)',
   'brand',
   `"instructor": ${instructorField}`,
   `'genre': ${genreField}`,
   'status',
-  "'slug' : slug.current",
-  "'permission_id': permission_v2",
+  '\'slug\' : slug.current',
```
adrian-musora: what's the reasoning behind switching to escaped strings? If we wanted a single format we should use ticks instead `

adrian-musora (approved):

PR #967, opened 2026-05-13
chore(sanity): typescript decorators

alesevero:
## Description/Design

- Refactors our need access and page type decorators:
  - Type-safety with typescript
  - Strict interface for base and specialized decorators
  - Base decorator has a much simpler logic, content needs to be in the correct shape (not if-else extravaganza)
  - Single walk decorator pipeline
  - Extract unnecessary logic up to the caller. Decorators decorate based on explicit inputs.

## Next steps

- Allow async decorators: #968


Meowbox64 (approved):
fantastic

PR #968, opened 2026-05-13
chore(sanity): async content decorators

alesevero:
## Jira Ticket(s)/RelatedPR(s)

- Built on top of #967

## Description/Design

- Enables the ability to run async content decorators in sequence or parallel

PR #969, opened 2026-05-13
chore(sanity): content navigate to decorator

alesevero:

- Built on top of #968 

## Description/Design

- Refactors navigateTo decorator to typescript

### DIFF
--- a/src/lib/sanity/decorators/base.ts
+++ b/src/lib/sanity/decorators/base.ts
@@ -7,11 +7,7 @@ export type DecorateFn<T, V> = (item: T) => V
 
 export type DecorateFnAsync<T, V> = (item: T) => Promise<V>
 
-export interface FieldDecorator<
-  T extends Decoratable,
-  K extends string = string,
-  V = unknown,
-> {
+export interface FieldDecorator<T extends Decoratable, K extends string = string, V = unknown> {
   field: K
   compute: DecorateFn<T, V>
   recurse?: boolean
@@ -142,7 +138,5 @@ async function visitAsync<T extends Decoratable>(
   if (!Array.isArray(children)) return
   const recursing = decorators.filter((d) => d.recurse !== false)
   if (recursing.length === 0) return
-  await Promise.all(
-    children.map((child) => visitAsync(child as T, recursing, depth + 1))
-  )
+  await Promise.all(children.map((child) => visitAsync(child as T, recursing, depth + 1)))
 }
--- a/src/lib/sanity/decorators/base.ts
+++ b/src/lib/sanity/decorators/base.ts
@@ -5,9 +5,16 @@ export interface Decoratable {
 
 export type DecorateFn<T extends Decoratable, V> = (item: T) => V
 
+export type DecorateFnAsync<T extends Decoratable, V> = (item: T) => Promise<V>
+
 export interface FieldDecorator<T extends Decoratable, V = unknown> {
   field: string
   compute: DecorateFn<T, V>
+}
+
+export interface FieldDecoratorAsync<T extends Decoratable, V = unknown> {
+  field: string
+  compute: DecorateFnAsync<T, V>
 }
 
 const MAX_CHILD_DEPTH = 3
@@ -45,4 +52,41 @@ function visit<T extends Decoratable>(
   for (const child of children) {
     visit(child as T, decorators, depth + 1)
   }
+}
+
+export function decorateAsync<T extends Decoratable, V>(
+  items: T | T[],
+  field: string,
+  compute: DecorateFnAsync<T, V>
+): Promise<T | T[]> {
+  return decorateAllAsync(items, [{ field, compute }])
+}
+
+export async function decorateAllAsync<T extends Decoratable>(
+  items: T | T[],
+  decorators: ReadonlyArray<FieldDecoratorAsync<T>>
+): Promise<T | T[]> {
+  const list = Array.isArray(items) ? items : [items]
+  await Promise.all(list.map((item) => visitAsync(item, decorators, 0)))
+  return items
+}
+
+async function visitAsync<T extends Decoratable>(
+  item: T,
+  decorators: ReadonlyArray<FieldDecoratorAsync<T>>,
+  depth: number
+): Promise<void> {
+  if (!item) return
+  const target = item as Record<string, unknown>
+  await Promise.all(
+    decorators.map(async ({ field, compute }) => {
+      target[field] = await compute(item)
+    })
+  )
+  if (depth >= MAX_CHILD_DEPTH - 1) return
+  const children = item.children
+  if (!Array.isArray(children)) return
+  await Promise.all(
+    children.map((child) => visitAsync(child as T, decorators, depth + 1))
+  )
 }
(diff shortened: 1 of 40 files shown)


src/lib/sanity/decorators/navigate-to.ts
```diff
@@ -0,0 +1,137 @@
...
+  findIncompleteLesson,
+  getLastInteractedOf,
+  getProgressState,
+  getProgressStateByIds,
+} from '../../../services/contentProgress.js'
+import { COLLECTION_TYPE } from '../../../services/sync/models/ContentProgress'
+
+export const NAVIGATE_TO_FIELD = 'navigate_to' as const
+
+const STATE_STARTED = 'started'
```
Meowbox64:
```suggestion
const STATE_STARTED = STATE.STARTED
```

src/lib/sanity/decorators/navigate-to.ts
```diff
@@ -0,0 +1,137 @@
+import { decorateAllAsync, type Decoratable, type FieldDecoratorAsync } from './base'
+import {
+  findIncompleteLesson,
+  getLastInteractedOf,
+  getProgressState,
+  getProgressStateByIds,
+} from '../../../services/contentProgress.js'
+import { COLLECTION_TYPE } from '../../../services/sync/models/ContentProgress'
+
+export const NAVIGATE_TO_FIELD = 'navigate_to' as const
```
Meowbox64: FE/MA are going to have to refactor if we change from `navigateTo`

src/lib/sanity/decorators/navigate-to.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export interface NavigateTo {
+  id: number
+  type: string
+  brand: string
+  thumbnail: string
+  published_on: string | null
+  status: string
+  child: NavigateTo | null
+  collection: { type: string; id: number } | null
```
Meowbox64: could use `CollectionParameter` from ContentProgress.ts
```suggestion
  collection: CollectionParameter | null
```

src/lib/sanity/decorators/navigate-to.ts
```diff
@@ -0,0 +1,137 @@
...
+  'course',
+  'guided-course',
+  'course-collection',
+  'song-tutorial',
+  'learning-path-v2',
+  'skill-pack',
+] as const
+
+const COURSE_FLOW_TYPES = ['course', 'skill-pack', 'song-tutorial']
+const GUIDED_FLOW_TYPES = ['guided-course', COLLECTION_TYPE.LEARNING_PATH]
```
Meowbox64, line 26: we should be careful about using this for LPs, since they have custom navigate_to logic if the LP is active (due to daily sessions). I'll think about this and let you know.

src/lib/sanity/decorators/navigate-to.ts
```diff
@@ -0,0 +1,137 @@
...
+  status: string
+  child: NavigateTo | null
+  collection: { type: string; id: number } | null
+}
+
+export type WithNavigateTo<T extends NavigateToDecoratable> = T & {
+  navigate_to: NavigateTo | null
+}
+
+function toNavigateTo(
```
Meowbox64: this name seems off to me. `buildNavigateTo`?

src/lib/sanity/decorators/base.ts
```diff
@@ -0,0 +1,148 @@
+export interface Decoratable {
+  children?: Decoratable[]
+  [key: string]: unknown
+}
+
+export type DecorateFn<T, V> = (item: T) => V
```
Meowbox64, line 6: why do these no longer have `<T extends Decoratable, V>`?

src/lib/sanity/decorators/base.ts
```diff
@@ -0,0 +1,148 @@
...
+  decorators: ReadonlyArray<FieldDecorator<T>>
+): T[]
+export function decorateAll<T extends Decoratable>(
+  items: T,
+  decorators: ReadonlyArray<FieldDecorator<T>>
+): T
+export function decorateAll<T extends Decoratable>(
+  items: T | T[],
+  decorators: ReadonlyArray<FieldDecorator<T>>
+): T | T[] {
```
Meowbox64, line 61: the return types for decorate and decorateAll have diverged. As far as I can see, these two fn's are just different parameter shapes for doing the same thing, so they should have the same response, no?

(I also noticed that compared to `chore/async-content-decorators`, `decorate` has since been changed from a wrapper of `decorateAll` to reformat parameters, to directly call `visit`)

Meowbox64 (approved):
excellent!

some comments, mostly my questions/observations.

src/lib/sanity/decorators/navigate-to.ts
alesevero: Yeah I was thinking about it but decided to leave it like this since the existing `getNavigateTo` has the same checks

src/lib/sanity/decorators/base.ts
alesevero: The constraint on `DecorateFn<T extends Decoratable, V>` is redundant at the fn type level. The constraint lives in the traversal fn: `FieldDecorator<T extends Decoratable>` and subsequent functions (`decorate`, `visit`). Constraint propagates from `FieldDecorator → DecorateFn` instantiation anyway (FieldDecorator's T is already Decoratable, so its compute: `DecorateFn<T,V>` resolves with constrained T.

The T on DecorateFn is not the same T on FieldDecorator. They are, but aren't. I know it's confusing 😅 

On `DecorateFn<T, V>`, T represents the type being passed down to `DecorateFn`. In this case, when DecorateFn is called, T is what the caller's type is, which already has the Decoratable constraint. If we check the `decorate` usage on `decorateAccess`:

image
image


the type signature of `decorate` becomes `decorate<AccessDecoratable, boolean>`. `AccessDecoratable` extends `Decoratable`.

src/lib/sanity/decorators/base.ts
alesevero: Yes, this PR contains changes on files from the other PRs because I started doing improvements here. That's my fault on poor change/pr management 😅 

The issue on return types is trickier, and I had to ask claude a few times to understand why we couldn't have the decorated type on `decorateAll`. Here is a summary of its answer:

`decorate` tracks single K/V in generics → return type `T & {[P in K]: V}` precise.

`decorateAll` takes `ReadonlyArray<FieldDecorator<T>>` — heterogeneous decorators, each with own K/V erased to defaults (K=string, V=unknown). No way to express union of added fields without variadic tuple gymnastics on input.

Options were:
- Lie: return `T & {[k:string]: unknown}` — pollutes index signature, kills type safety on existing fields.
- Tuple-infer: `<D extends readonly FieldDecorator<T, any, any>[]>` then map — 
  - works but callers must pass as const and inference gets fragile with recurse/optional fields.
- Honest: return T. Caller knows fields added but TS doesn't track. 
  - Trade precision for ergonomics.

Picked honest. decorate keeps precise typing for single-field common case; decorateAll punts on multi-field tracking.

We can still do `decorateAll(...) as DecoratedTypeOfOurChoice` and that will allow TS to infer the types on whatever the function returned.

I'll eventually find a way to fix this. As of now, it works.

src/lib/sanity/decorators/navigate-to.ts
alesevero: Good catch! I'll actually remove `STATE_STARTED` and just use them enum directly

PR #930, opened 2026-04-20
feat(TP-1195): use client-side filtering for live events to leverage sanity cache

sheldonbaker:
## Jira Ticket(s)/RelatedPR(s)

[TP-1195](https://musora.atlassian.net/browse/TP-1195)

## Description/Design

Fetches all live streams after start of current day, and filters records client-side.

Still need to confirm expected new cache behavior, and verify that this query strategy works with it.

## Testing

[TP-1195]: https://musora.atlassian.net/browse/TP-1195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #971, opened 2026-05-13
feat(TP-1195): optimize live event loading

sheldonbaker:
Improvement off of #930.

Adds an upper bound to sanity query, and changes filter to be start-of-yesterday to end-of-tomorrow (UTC) to avoid edge cases around when clock passes midnight

Will change base to `main` once re-approved

### DIFF
--- a/src/services/sanity.js
+++ b/src/services/sanity.js
@@ -1216,7 +1216,7 @@ export async function fetchLiveEvent(brand, forcedContentId = null) {
   const filter = await new FilterBuilder(baseFilter, { bypassPermissions: true }).buildFilter()
 
   const events = await fetchSanity(
-    `*[${filter}]{${liveEventFields}} | order(live_event_start_time asc)`,
+    `*[${filter}]{${fieldsString}} | order(live_event_start_time asc)`,
     true,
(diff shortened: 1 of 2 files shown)


src/services/sanity.js
```diff
@@ -1203,15 +1201,25 @@ export async function fetchLiveEvent(brand, forcedContentId = null) {
...
-      && live_event_end_time >= '${getSanityDate(endDateTemp, false)}'`
+      && live_event_start_time >= '${startOfYesterday}'
+      && live_event_start_time < '${endOfTomorrow}'`
 
   const filter = await new FilterBuilder(baseFilter, { bypassPermissions: true }).buildFilter()
 
-  // This query finds the first scheduled event (sorted by start_time) that ends after now()
-  const query = `*[${filter}]{${fieldsString}} | order(live_event_start_time)[0...1]`
+  const events = await fetchSanity(
+    `*[${filter}]{${liveEventFields}} | order(live_event_start_time asc)`,
```
Meowbox64: this will break if we dont use `fieldsString`

Meowbox64 (approved):

PR #974, opened 2026-05-14
feat(BEHSTP-333): convert profile js to ts and add updateProfileVisibility method

roxanariza:

- [BEHSTP-333](https://musora.atlassian.net/browse/BEHSTP-333)
- [BEHSTP-334](https://musora.atlassian.net/browse/BEHSTP-334)

## Description/Design

- Converted Profile.js to Profile.ts
- New method updateProfileVisibility  to set profile public or private
- New MCS method getCompletedAwardsByUser

## Testing

- _How can this be verified?_
- _Setup required_
- _what page(s) are applicable_
- _what action(s) should a user take_
- _what is the expected outcome?_

## Additional Notes

- I will create tests for the new method getCompletedAwardsByUser

[BEHSTP-333]: https://musora.atlassian.net/browse/BEHSTP-333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[BEHSTP-334]: https://musora.atlassian.net/browse/BEHSTP-334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/services/user/profile.ts
```diff
@@ -0,0 +1,66 @@
...
+    total_practice_time: longestStreaks.totalPracticeSeconds + (stats.v1_practice_time ?? 0),
+  } as OtherStatsDTO
+}
+
+export async function deleteProfilePicture(): Promise<void> {
+  const url = `${baseUrl}/v1/users/profile_picture`
+  await DELETE(url)
+}
+
+export async function updateProfileVisibility(isPublic: boolean): Promise<UserResource> {
```
adrian-musora, line 62: why is this paramaterized with the user?
Similar to other user update endpoints, users (or potentially admins) should be able to edit users. `deleteProfilePicture` does this cleanly.
If we're allow admins to edit, then the endpoint needs to take the user id as an explicit input, if we're not allowing admins to view, we should remove the parameter entirely.

roxanariza: The admins will not be able to update user profile visibility. I updated the endpoint and the MCS method

sheldonbaker (approved):

PR #975, opened 2026-05-14
fix: do immediate push after record repair

sheldonbaker:
Does an immediate push of repaired records, for #963.

Easier debugging/verification in production.

Tested push call locally.


slumrig (approved):
👍

PR #976, opened 2026-05-15
refactor: content progress typescript part 1

alesevero:
Built on top of #969 

Part two: #978 (you might as well just review part 2 if you want as it adds a significant structural change to the progress module)

## Summary
- New TypeScript progress module `src/services/progress/index.ts` exposing state / resume-time / last-interacted queries plus `findIncompleteLesson`, consumed by the `navigate-to` decorator
- `decorateNavigateTo` pre-fetches parent + child progress states in a single `getProgressStateByIds` batch, threaded into `computeNavigateTo` via a context object — drops ~2N round-trips on N-item lists
- Decorator usage reference moved from `src/lib/sanity/decorators/examples.ts` to `usage/sanity/decorators/examples.ts` so it stops shipping in the published package
- Reduced function calls by removing most of the normalizing functions as we now can enforce types at development time.

## Additional Notes

- This cleans up the interface as well. There were some functions that did the same thing but filtered out the data coming back, doing unnecessary traversals and widening the module's interface.

## Test plan
- `npx jest --no-coverage` — 56 suites, 957 passing
- Verify decorator usage in downstream consumers once wired up
- Manual smoke of navigate-to behavior on course / course-collection / learning-path content

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,44 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+### [2.160.4](https://github.com/railroadmedia/musora-content-services/compare/v2.160.3...v2.160.4) (2026-05-20)
+
+
+### Bug Fixes
+
+* **MU2-1510:** process need access for live events ([#964](https://github.com/railroadmedia/musora-content-services/issues/964)) ([66aa2b5](https://github.com/railroadmedia/musora-content-services/commit/66aa2b596b28c272b938672522b9c4024f9e328d))
+* **MU2-1511:** add need access to upcoming live events ([#965](https://github.com/railroadmedia/musora-content-services/issues/965)) ([9366907](https://github.com/railroadmedia/musora-content-services/commit/936690752529d25d80df0192e80b57ad90921445))
+* **MU2-1512:** add child data to new and upcoming ([#960](https://github.com/railroadmedia/musora-content-services/issues/960)) ([329f5c8](https://github.com/railroadmedia/musora-content-services/commit/329f5c8b98ea11f1d456dd53e81b705f55b4f346))
+
+### [2.160.3](https://github.com/railroadmedia/musora-content-services/compare/v2.160.2...v2.160.3) (2026-05-14)
+
+
+### Bug Fixes
+
+* do immediate push after record repair ([#975](https://github.com/railroadmedia/musora-content-services/issues/975)) ([c087536](https://github.com/railroadmedia/musora-content-services/commit/c08753659fb5b1342ae8499fbf296ed4a2197625))
+
+### [2.160.2](https://github.com/railroadmedia/musora-content-services/compare/v2.160.1...v2.160.2) (2026-05-14)
+
+
+### Bug Fixes
+
+* **BR-678:** fix issues with is_locked and collection ([#972](https://github.com/railroadmedia/musora-content-services/issues/972)) ([725529b](https://github.com/railroadmedia/musora-content-services/commit/725529bd69d05ab79113b21a5e3dca5ce1deb0dc))
+
+### [2.160.1](https://github.com/railroadmedia/musora-content-services/compare/v2.160.0...v2.160.1) (2026-05-13)
+
+
+### Bug Fixes
+
+* disable debug code ([#970](https://github.com/railroadmedia/musora-content-services/issues/970)) ([4046c4f](https://github.com/railroadmedia/musora-content-services/commit/4046c4fca4457b9d21eab5221876a991bdf22973))
+
+## [2.160.0](https://github.com/railroadmedia/musora-content-services/compare/v2.159.0...v2.160.0) (2026-05-13)
+
+
+### Features
+
+* adds stale record cleanup ([#963](https://github.com/railroadmedia/musora-content-services/issues/963)) ([72d917f](https://github.com/railroadmedia/musora-content-services/commit/72d917ff0d64d8c27f1d4ec831233c17663b1c49))
+* use client-side filtering for live events to leverage sanity cache ([#930](https://github.com/railroadmedia/musora-content-services/issues/930)) ([cbfe024](https://github.com/railroadmedia/musora-content-services/commit/cbfe02476b654a240bb1ba35093a98a03535227c))
+
 ## [2.159.0](https://github.com/railroadmedia/musora-content-services/compare/v2.158.3...v2.159.0) (2026-05-12)
 
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "musora-content-services",
-  "version": "2.159.0",
+  "version": "2.160.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "musora-content-services",
-      "version": "2.159.0",
+      "version": "2.160.4",
       "license": "ISC",
       "dependencies": {
         "@sanity/client": "^5.4.2",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "musora-content-services",
-  "version": "2.159.0",
+  "version": "2.160.4",
   "description": "A package for Musoras content services ",
   "main": "src/index.js",
   "type": "module",
--- a/src/contentTypeConfig.js
+++ b/src/contentTypeConfig.js
@@ -698,6 +698,7 @@ export let contentTypeConfig = {
       'show_in_new_feed',
       isLiveField(),
     ],
+    includeChildFields: true,
   },
 }
 
--- a/src/infrastructure/sanity/README.md
+++ b/src/infrastructure/sanity/README.md
@@ -1,0 +1,230 @@
+# Sanity Infrastructure
+
+This module provides a TypeScript-based infrastructure for interacting with Sanity CMS, following the same architectural patterns as the HTTP client.
+
+## Architecture
+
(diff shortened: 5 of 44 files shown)


alesevero:
## Automated PR Review 🤖

### Context

- **Ticket:** None found
- **Modules:** src/lib/sanity/decorators, src/services/progress, src/services/contentProgress
- **Commits:** 12
- **Agents run:** style, logic, security, test-coverage
- **Skipped:** database (JS/TS project, no migrations/Models), functional (no Jira ticket ID found)

---

### Complexity: Large

~2,139 lines changed across 12 files. Mixes three concerns: new Sanity decorator framework (`base`/`need-access`/`page-type`/`navigate-to`), new TS progress module (`src/services/progress/index.ts`), and a large Prettier-style reformat of `src/services/contentProgress.js` (+182 / -149, ~90% formatting). Risk concentrated in the navigate-to decorator and progress module due to behavior shifts vs. the legacy JS counterparts.

---

### Issues

#### Logic / Behavior

- `src/lib/sanity/decorators/navigate-to.ts:478-487` — ⚠️ Behavior regression: `getLastInteractedOf` may resolve to `undefined`. `childrenStates.get(undefined)` is `undefined`, then `findIncompleteLesson(..., undefined, ...)` does `Number(undefined) → NaN → indexOf === -1 → null`, then `childrenById.get(null) → undefined → null`. STARTED parent with no child progress now returns `null`; the legacy JS flow fell back to `ids[0]` (first child). Fix: when `lastInteractedId` is `undefined`, fall through to the non-STARTED branch (first-child fallback).
- `src/lib/sanity/decorators/navigate-to.ts:498-500` — ⚠️ TWO_DEPTH STARTED branch has the same `undefined` issue: `childrenById.get(undefined)` → `null` for a STARTED `course-collection` with no child progress, even though non-STARTED falls back to the first child. Fix: default `lastInteractedId` to `children[0].id` for the TWO_DEPTH STARTED branch.
- `src/lib/sanity/decorators/navigate-to.ts:487` — ⚠️ When `findIncompleteLesson` returns `null` (e.g. stale `lastInteracted` not in current children), `target` is `undefined` and the function returns `null`. Legacy JS always returned an id. Fix: fall back to `children[0].id` when `targetId` is null/undefined.
- `src/services/progress/index.ts:33 (findIncompleteLesson)` — ⚠️ When `currentIndex === -1`, returns `null`. Legacy JS in `contentProgress.js` returned `ids[0]` as fallback. This is the root cause of the navigate-to regressions above. Either restore the JS behavior or document the new contract and audit all call sites.
- `src/lib/sanity/decorators/navigate-to.ts:416` — `NAVIGABLE_TYPES` hardcodes `'learning-path-v2'` while `GUIDED_FLOW_TYPES` uses `COLLECTION_TYPE.LEARNING_PATH`. If the enum value ever changes, the two silently diverge. Fix: replace the literal with `COLLECTION_TYPE.LEARNING_PATH`.
- `src/services/progress/index.ts (getAllStartedOrCompleted)` — ⚠️ The `onlyIds` option declared on `GetAllQueryOptions` is silently ignored — function always returns full records. Legacy JS branched on `onlyIds: true` to return `data.map(rec => rec.content_id)`. Callers relying on this get materially larger payloads. Fix: honor `onlyIds` or remove it from the type.
- `src/services/progress/index.ts (getAllStartedOrCompleted)` — Magic `60 * 24 * 60 * 60` (60 days in seconds) is hardcoded with no comment. Legacy JS had `// 60 days in seconds` annotation, which was dropped. Fix: extract `const SIXTY_DAYS_IN_SECONDS = 60 * 24 * 60 * 60`.
- `src/services/progress/index.ts:23 (getLastInteractedOf)` — Return type implicit `number | undefined`, but call sites accept it as `number`. Fix: declare explicit `Promise<number | undefined>` and update callers.
- `src/services/progress/index.ts (getResumeTimeSecondsByIds)` — `parseInt(r.data)` missing radix. Use `parseInt(r.data, 10)` or `Number(r.data)`.
- `src/services/progress/index.ts (findIncompleteLesson)` — Duplicated between `progress/index.ts` and `contentProgress.js` with divergent signatures (Map+object vs Map only). Pick one source of truth or have one delegate.

#### Style / Conventions

- `src/lib/sanity/decorators/base.ts:78,83,85,87,135,142,144,146` — Single-line `if`/`return` without curly braces. Codebase convention is braces on all control structures.
- `src/lib/sanity/decorators/navigate-to.ts:463,466,500` — Same: missing curly braces on single-line `if`.
- `src/lib/sanity/decorators/navigate-to.ts:463` — `content.type as (typeof NAVIGABLE_TYPES)[number]` cast defeats the membership-check intent. Fix: `(NAVIGABLE_TYPES as readonly string[]).includes(content.type)`.
- `src/lib/sanity/decorators/navigate-to.ts:416` — `COURSE_FLOW_TYPES`/`GUIDED_FLOW_TYPES`/`TWO_DEPTH_TYPES` are plain arrays without `as const`, while `NAVIGABLE_TYPES` is `as const`. Make consistent.
- `src/lib/sanity/decorators/page-type.ts:608` — `item.type as string` cast hides the `undefined` case. Fix: `SONG_TYPES_WITH_CHILDREN.includes(item.type ?? '')`.
- `src/services/progress/index.ts` — Top-level exports declared as `export const foo = async (...) =>` arrow expressions; the rest of the codebase uses `export async function`. `tools/generate-index.cjs` scans for `export function` / `export async function` — arrow exports may be missed by the index generator. Fix: use `export async function` consistently.
- `src/services/progress/index.ts:1299` — `collection = null` default is untyped (inferred `null`). Use `collection?: CollectionParameter` matching sibling functions.
- `src/services/progress/index.ts:1349,1354` — `limit = null` defaults untyped; declare `number | null`.
- `src/services/progress/index.ts (getByIds/getById/getByRecordIds)` — `dataKey: string` allows arbitrary record indexing. Narrow to `keyof ProgressRecord` (or a known union) — tightens typing and removes the latent property-bleed risk if these helpers ever take external input.
- `src/lib/sanity/decorators/examples.ts` — Lives in `src/`, ~229 lines of example/doc code. `tools/generate-index.cjs` only scans `src/services/`, so it won't auto-export, but the file still ships in the published package. Fix: move to `docs/` or `examples/`.
- `src/services/contentProgress.js` — ~90% Prettier reformatting churn (trailing commas, arrow-fn parens, indentation) mixed with substantive PR. Split the formatting-only reflow into a separate PR/commit to keep this review tractable. PR body already acknowledges this was committed unintentionally.

#### Performance

- `src/lib/sanity/decorators/navigate-to.ts:462` — `computeNavigateTo` fires 3 async DB calls (`getProgressState`, `getProgressStateByIds`, `getLastInteractedOf`) per parent. Decorating N parents = ~3N parallel queries. Suggest a pre-pass that batches `getProgressStateByIds(parentIds)` and `getLastInteractedOf` across all items before per-item compute.
- `src/lib/sanity/decorators/navigate-to.ts:498` — TWO_DEPTH branch recursively re-calls `computeNavigateTo`, doubling per-item queries for `course-collection`. Thread already-fetched state through recursion.
- `src/lib/sanity/decorators/base.ts:137 (visitAsync)` — No cross-decorator or cross-node de-duplication. If two decorators independently fetch the same per-id data, work is duplicated across the tree. Document that async decorators must not perform per-item I/O, or provide a shared per-tree cache to `compute`.

#### Test Hygiene

- `test/unit/lib/sanity/decorators/navigate-to.test.ts:413-417` — Mocks `src/services/sanity.js` (`getHierarchy`, `getHierarchies`, `getSanityDate`), but `navigate-to.ts` does not import any of them. Dead mock.
- `test/unit/lib/sanity/decorators/navigate-to.test.ts:408-411` — Mocks `content-org/learning-paths`; not imported by the code under test. Dead mock.
- `test/unit/lib/sanity/decorators/navigate-to.test.ts:401-404` — `practices` mock (`queryAll`, `getAll`) defined but never invoked by `navigate-to.ts`. Dead mock.
- `test/unit/services/progress.test.ts:995` — `mockLastInteracted = '101' as any` bypasses the declared `number | null` type to test that `parseInt(r.data)` parses a string. This suggests the production return type for `r.data` is actually `string`, not `number | null`. Either fix the test typing or fix the production type to match reality.

---

### Test Coverage

**Proportionality: Adequate** — 5 new test files cover base decorators, need-access, page-type, navigate-to, and the new progress module.

**Gaps:**
- `decorateNavigateTo` — "course/guided-course STARTED, but `lastInteracted` id is not in `childrenById`" is not tested for the single-depth flows (only TWO_DEPTH is covered). Given the behavior regressions flagged in the Logic section, this is exactly the path that needs explicit coverage — add tests with a STARTED parent whose `mockLastInteracted` points outside its children and assert expected fallback (first child or first incomplete).
- `decorateNavigateTo` — No positive-case test exercises a non-null `collection` param through `buildNavigateTo`. Minor since current callers pass `null`.
- `getAllStartedOrCompleted` — Once `onlyIds` is fixed (or removed), add a test asserting the chosen contract.

---

### Notes

- Codecov reports 75.78% patch coverage — driven entirely by `examples.ts` being uncovered (0%). Once `examples.ts` is moved out of `src/`, the patch number should rise significantly.
- PR body acknowledges the `contentProgress.js` formatting changes were committed by accident. Strongly recommend splitting that out before merge — it makes the diff substantially easier to review and bisect.
- The duplication of `findIncompleteLesson` (JS vs TS) is fine as a transitional state for "part 1", but track removal of the JS version in a follow-up so the two don't drift further.

---

_Automated review — see `.claude/skills/pr-review/SKILL.md`_

sheldonbaker (approved):
Pending nitpicks in https://github.com/railroadmedia/musora-content-services/pull/985

PR #978, opened 2026-05-19
refactor: content progress typescript part 2

alesevero:
## Jira Ticket(s)/Related PR(s)

- Built from #976 

## Description/Design

- The progress module was a single flat `index.ts` — split into `state.ts`, `collections.ts`, `types.ts`, and `internal/queries.ts` to separate read queries by concern and keep private DB helpers out of the public surface
- All functions are now exposed via a `Progress` namespace object, giving callers unambiguous import paths and a consistent naming convention (nouns for reads; verbs reserved for future write operations)
- `navigate-to.ts` updated to consume `Progress.*` and to match the updated `incompleteLesson` parameter order (`contentType` before `currentContentId`)

## Testing

- How to verify: run `npm test test/unit/services/progress` and `npm test test/unit/lib/sanity/decorators/navigate-to`
- Environment: local
- Expected outcome: all tests pass, no TypeScript errors from `npx tsc --noEmit`

## Additional Notes

- Write operations (`recordWatchSession`, `saveContentProgress`, etc.) are not yet migrated — they remain in `contentProgress.js` and will land in a follow-up PR
- The `Progress` namespace pattern is intentional: avoids naming collisions with the legacy JS exports during the transition period


src/services/progress/state.ts
```diff
@@ -0,0 +1,50 @@
...
+): number | null | undefined => {
+  const ids = Array.from(progressOnItems.keys())
+  const getProgress = (id: number) => progressOnItems.get(id)
+
+  if (contentType === 'guided-course' || contentType === COLLECTION_TYPE.LEARNING_PATH) {
+    return ids.find((id) => getProgress(id) !== 'completed') || ids.at(0)
+  }
+
+  const currentIndex = ids.indexOf(Number(currentContentId))
+  const startIndex = currentIndex === -1 ? 0 : currentIndex + 1
```
sheldonbaker, line 40: Old behavior seemed to return `null` if `currentIndex` was `NaN`. New behaviour will instead search through ids for first incomplete lesson.

Intentional?

src/lib/sanity/decorators/navigate-to.ts
sheldonbaker: Not from this PR directly, but this decorator doesn't have the same batching that `decorateNavigateTo` does (multiple DB round trips). Is this decorator meant for public consumption?

src/lib/sanity/decorators/navigate-to.ts
```diff
@@ -113,21 +108,21 @@ async function computeNavigateTo(
   const childrenById = new Map(children.map((c) => [c.id, c]))
   const childrenStates = ctx
     ? new Map(childrenIds.map((id) => [id, ctx.states.get(id) ?? '']))
-    : await getProgressStateByIds(childrenIds)
-  const lastInteractedId = await getLastInteractedOf(childrenIds)
+    : await Progress.stateByIds(childrenIds)
+  const lastInteractedId = await Progress.lastInteractedOf(childrenIds)
```
sheldonbaker, line 112: This makes N+1 queries - could a single `lastInteractedOf` call be batched earlier and passed into `ctx` just like is done for `Progress.state`?

src/lib/sanity/decorators/navigate-to.ts
alesevero: Good catch, it should be a factory function similar to accessDecorator

src/services/progress/state.ts
alesevero: I think it's better behaviour. _No last-interacted_ should navigate to first incomplete, not return null. The old null return was probably an oversight? Worth confirming, tho.

alesevero: Also now that we have typing and more tests, devs should be able to catch inconsistencies when using those functions?

src/lib/sanity/decorators/navigate-to.ts
alesevero: Yeah that's a good point. Old navigate-to decorator has the same issue. Ran by claude and the optimization wouldn't reduce the number of queries, but make them parallel. Could just increase complexity.

Might be worth as a separate improvement

sheldonbaker (approved):

PR #979, opened 2026-05-19
refactor: content progress typescript part 3

alesevero:
## Jira Ticket(s)/Related PR(s)

- Stacked on top of #978 (content progress TS part 2)

## Description/Design

- Migrated the remaining read functions from `contentProgress.js` into `progress/state.ts` and `progress/collections.ts`: `snapshotByIds`, `snapshotByRecordIds`, `methodAccessedIds`, `percentByContentId` — exposed through the `Progress` namespace
- Created `progress/utils.ts` with `generateRecordId` and `extractFromRecordId`; normalization helpers were dropped because TypeScript enforces the input types at compile time
- Created `progress/internal/bubble.ts` with the pure bubble/trickle helpers and the safe write wrapper (`bubbleAndTrickleProgressesSafely`); LP duplication chain stays out of `bubble.ts` to avoid a circular dependency with `saveContentProgress` (will land with mutations in a follow-up)
- Added `ProgressSnapshot`, `ProgressMetadata`, `Hierarchy`, and `RecordIdParts` to `types.ts` and re-exported them from `index.ts`

## Testing

- How to verify: `npm test test/unit/services/progress` and `npm test test/unit/services/progress-internal`
- Environment: local
- Expected outcome: all new test suites pass, no TypeScript errors from `npx tsc --noEmit`

## Additional Notes

- Watch-session writes (`sessions.ts`) and status writes (`mutations.ts`) are not in this PR — they depend on the bubble internals merged here and on each other, so they will land in a follow-up that also removes `contentProgress.js`


src/services/progress/internal/bubble.ts
```diff
@@ -0,0 +1,141 @@
...
+  collection?: CollectionParameter,
+  { bubble = true, trickle = true }: { bubble?: boolean; trickle?: boolean } = {}
+): Promise<Record<number, number>> => ({
+  ...(trickle ? trickleProgress(hierarchy, contentId, progress) : {}),
+  ...(bubble ? await bubbleProgress(hierarchy, contentId, collection) : {}),
+})
+
+export interface BubbleAndTrickleOptions {
+  isResetAction?: boolean
+  accessedDirectly?: boolean
```
Meowbox64, line 105: this needs to be default true, or else refactor callers.

src/services/progress/collections.ts
```diff
@@ -31,3 +31,11 @@ export const allStartedOrCompleted = async (
       updatedAfter: Math.floor(Date.now() / 1000) - SIXTY_DAYS_IN_SECONDS,
     })
     .then((r) => r.data)
+
+export const percentByContentId = async (brand?: string): Promise<Record<number, number>> =>
```
Meowbox64: what is this fn for? its not called anywhere currently. 

also consider renaming. we use "ByContentId" to indicate content id is the parameter, but here its indicating it returns a content_id->progress_percent map. took me a few minutes to understand what was happening because the fn name was misleading.

src/services/progress/state.ts
```diff
@@ -48,3 +49,49 @@ export const incompleteLesson = (
...
+  contentIds: number[],
+  collection?: CollectionParameter
+): Promise<Record<number, ProgressSnapshot>> => {
+  const result = Object.fromEntries(
+    contentIds.map((id) => [id, { last_update: 0, progress: 0, status: '' }])
+  ) as Record<number, ProgressSnapshot>
+
+  await db.contentProgress.getSomeProgressByContentIds(contentIds, collection).then((r) => {
+    r.data.forEach((p) => {
+      result[p.content_id] = {
```
Meowbox64: beware returning a number-keyed js object. It reorders the keys numerically, which has given me many headaches in the past. returning a map may be a better idea.

src/services/progress/state.ts
```diff
@@ -48,3 +49,49 @@ export const incompleteLesson = (
...
+  ids: string[]
+): Promise<Record<string, ProgressSnapshot>> => {
+  const result = Object.fromEntries(
+    ids.map((id) => [id, { last_update: 0, progress: 0, status: '' }])
+  ) as Record<string, ProgressSnapshot>
+
+  await db.contentProgress.getSomeProgressByRecordIds(ids).then((r) => {
+    r.data.forEach((p) => {
+      result[p.id] = {
+        last_update: p.updated_at,
```
Meowbox64, line 84: this fn and the above fn do the same thing, but return different data for `last_update`. This is my fault, and is related to homepage progress row needs. lets chat about this for what we should do.

Meowbox64 (approved):
fantastic :chef-kiss:

PR #982, opened 2026-05-20
fix(live-events): fetch permission ids

alesevero:

- N/A

## Description/Design

- Add `permission_id` to the minimum live event fields returned by `getLiveFields()` so live event consumers can resolve access without an extra query.

## Testing

- How to verify: call any live event query path that uses `getLiveFields()` and confirm `permission_id` is present on the returned payload.
- Environment: local
- Expected outcome: live event responses include `permission_id` alongside existing live fields.

## Additional Notes

- Remaining diff in `contentTypeConfig.js` is incidental Prettier-style formatting (quote and trailing-comma normalization).


sheldonbaker (approved):
✋

PR #992, opened 2026-06-03
fix(BR-654): dont set resume-time to null

sheldonbaker:
[BR-654](https://musora.atlassian.net/browse/BR-654)

`null` was being passed to `saveContentProgress`, which means something different than `undefined`. Passing `undefined` instead will not alter resume time at all.

[BR-654]: https://musora.atlassian.net/browse/BR-654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Meowbox64 (approved):
oops haha, nice catch

PR #993, opened 2026-06-04
feat: add need access logic branch for LPs

Meowbox64:

- LPs are a special case where they have no content permissions, but their children do. Their children are not "owned" by the LP but "borrowed" from skill packs and song tutorials. his means that our current need_access decorator reads the LP as having no permissions and therefore sets `need_access` as `false`, even if all lessons within it get set as `need_access=true`.

## Description/Design

- updates needAccessDecorator to run on children first (bottom of hierarchy up), and use children `need_access` values to determine parent's `need_access` if its an LP.

## Testing

- use `staging` sanity dataset, head to FE devendpoint. (or use BE mcs-call)
- call `fetchByRailContentIds` with `[452387], 'learning-path-v2'`
- call `fetchByRailContentId` with `452387, 'learning-path-v2'`
- both should set `need_access: true` on the parent. (and all children should be too)

## Additional Notes

- we could optionally remove the type check, since this is likely code we'd want run on all types.

### DIFF
--- a/src/services/permissions/PermissionsV2Adapter.ts
+++ b/src/services/permissions/PermissionsV2Adapter.ts
@@ -55,6 +55,9 @@ export class PermissionsV2Adapter extends PermissionsAdapter {
 
     // Content with no permissions is accessible to all
     if (contentPermissions.size === 0) {
+      if (content?.type === 'learning-path-v2' && content?.children?.length) {
+        return content.children.every(child => child.need_access)
+      }
       return false
     }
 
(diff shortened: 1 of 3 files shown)


src/services/permissions/PermissionsV2Adapter.ts
```diff
@@ -55,6 +55,9 @@ export class PermissionsV2Adapter extends PermissionsAdapter {
 
     // Content with no permissions is accessible to all
     if (contentPermissions.size === 0) {
+      if (content?.type === 'learning-path-v2' && content?.children?.length) {
```
eprachnau: is there an enum we can use for this?

Meowbox64: yes

eprachnau (approved):
👌

PR #994, opened 2026-06-08
feat(BEHSTP-364): User video player settings

roxanariza:

- [ BEHSTP-365](https://musora.atlassian.net/browse/BEHSTP-365)
- [MPB PR](https://github.com/railroadmedia/musora-platform-backend/pull/1230)

## Description/Design

  - Added fetchPlayerSettings and updatePlayerSettings functions to fetch and partially update the user's video player settings (auto_play, auto_next, auto_complete, playlist_auto_next) via the MPB API.
  - Added PlayerSettings and UpdatePlayerSettingsData TypeScript interfaces to types.d.ts                                                                                                                                          
  - Exported both functions from index.js / index.d.ts                                                                                                                                                                             

## Testing

  - [ ] fetchPlayerSettings calls GET /api/user-management-system/v1/user/player-settings and returns the settings object.
  - [ ] updatePlayerSettings calls PATCH on the same endpoint with only the provided fields (partial update supported).
  - [ ] All 6 unit tests pass

## Additional Notes

- _Provide any additional context or notes that might be helpful for the reviewer._

### DIFF
--- a/test/unit/playerSettings.test.ts
+++ b/test/unit/playerSettings.test.ts
@@ -13,14 +13,14 @@ const mockPlayerSettings: PlayerSettings = {
 
 describe('playerSettings', () => {
   let mockGet: jest.Mock
-  let mockPut: jest.Mock
+  let mockPatch: jest.Mock
 
   beforeEach(() => {
     mockGet = jest.fn().mockResolvedValue(mockPlayerSettings)
-    mockPut = jest.fn().mockResolvedValue(mockPlayerSettings)
+    mockPatch = jest.fn().mockResolvedValue(mockPlayerSettings)
     ;(HttpClient as jest.Mock).mockImplementation(() => ({
       get: mockGet,
-      put: mockPut,
+      patch: mockPatch,
     }))
   })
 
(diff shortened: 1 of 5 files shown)


test/unit/playerSettings.test.ts
```diff
@@ -0,0 +1,73 @@
...
+        { playlist_auto_next: false }
+      )
+    })
+
+    test('sends all fields when all are provided', async () => {
+      const allFields = { auto_play: false, auto_next: false, auto_complete: false, playlist_auto_next: false }
+      await updatePlayerSettings(allFields)
+      expect(mockPatch).toHaveBeenCalledWith(expect.any(String), allFields)
+    })
+
```
Meowbox64, line 65: maybe we should add a test to verify that even if only 1 field is provided, all are returned; that doesn't exist here yet.

Meowbox64 (approved):

test/unit/playerSettings.test.ts
roxanariza: This is already covered in the 'returns the updated player settings from the API' test—updatePlayerSettings is called with only { auto_next: false } (1 field), but the mock resolves with all 4 fields and we assert the full object is returned. Since the endpoint behavior is mocked, we can't test the actual BE response shape here — that would require a live test.

PR #999, opened 2026-06-16
feat(BEHSTP-370): add need_lifetime_upgrade flag and grant30DaysAccessForLifetime

roxanariza:
## Summary
- Adds `need_lifetime_upgrade` flag on Sanity content documents — set to `true` for lifetime members who don't have Plus and encounter Plus-gated content they cannot access (used to show a Plus upsell)
- Adds `grant30DaysAccessForLifetime()` API method that calls `/v1/grant-30-days-access-for-lifetime`

## Test plan
- [ ] Verify `need_lifetime_upgrade` is `true` for a lifetime (non-Plus) member on Plus-gated content they can't access
- [ ] Verify `need_lifetime_upgrade` is `false` for admin, Plus members, non-lifetime members, or content the user already has access to
- [ ] Verify `grant30DaysAccessForLifetime()` hits the correct endpoint

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/lib/sanity/decorators/need-lifetime-upgrade.ts
+++ b/src/lib/sanity/decorators/need-lifetime-upgrade.ts
@@ -1,0 +1,49 @@
+import { decorate, type FieldDecorator } from './base'
+import { type UserPermissions } from '../../../services/permissions'
+import { type AccessDecoratable } from './need-access'
+import { MEMBERSHIP_PERMISSIONS } from '../../../constants/membership-permissions'
+
+export const NEED_LIFETIME_UPGRADE_FIELD = 'need_lifetime_upgrade' as const
+
+type LifetimeUpgradeDecoratable = AccessDecoratable & { need_access?: boolean }
+
+export type WithNeedLifetimeUpgrade<T extends AccessDecoratable> = T & {
+  need_lifetime_upgrade: boolean
+  children?: WithNeedLifetimeUpgrade<NonNullable<T['children']>[number]>[]
+}
+
+export function lifetimeUpgradeDecorator(
+  userPermissions: UserPermissions
+): FieldDecorator<LifetimeUpgradeDecoratable, typeof NEED_LIFETIME_UPGRADE_FIELD, boolean> {
+  const userPermSet = new Set(userPermissions?.permissions ?? [])
+  const hasLifetime = userPermSet.has(MEMBERSHIP_PERMISSIONS.lifetime)
+  const hasPlus = userPermSet.has(MEMBERSHIP_PERMISSIONS.plus)
+
+  return {
+    field: NEED_LIFETIME_UPGRADE_FIELD,
+    compute: (item) => {
+      if (userPermissions?.isAdmin || !hasLifetime || hasPlus) return false
+      if (!item.need_access) return false
+      const contentPerms = new Set(item?.permission_id ?? [])
+      return contentPerms.has(MEMBERSHIP_PERMISSIONS.plus)
+    },
+  }
+}
+
+export function decorateLifetimeUpgrade<T extends LifetimeUpgradeDecoratable>(
+  items: T[],
+  userPermissions: UserPermissions
+): WithNeedLifetimeUpgrade<T>[]
+export function decorateLifetimeUpgrade<T extends LifetimeUpgradeDecoratable>(
+  items: T,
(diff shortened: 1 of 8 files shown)


src/lib/sanity/decorators/need-lifetime-upgrade.ts
```diff
@@ -0,0 +1,49 @@
...
+): FieldDecorator<LifetimeUpgradeDecoratable, typeof NEED_LIFETIME_UPGRADE_FIELD, boolean> {
+  const userPermSet = new Set(userPermissions?.permissions ?? [])
+  const hasLifetime = userPermSet.has(MEMBERSHIP_PERMISSIONS.lifetime)
+  const hasPlus = userPermSet.has(MEMBERSHIP_PERMISSIONS.plus)
+
+  return {
+    field: NEED_LIFETIME_UPGRADE_FIELD,
+    compute: (item) => {
+      if (userPermissions?.isAdmin || !hasLifetime || hasPlus) return false
+      if (!item.need_access) return false
```
Meowbox64, line 26: Im not sure I like the idea of this decorator depending on another decorator (need_access) being run first, but maybe its ok. We need it for content owners

we could maybe check if any of the user's permissions overlap with the content permissions, but then this is just implementing `need_access` again lol

Meowbox64 (approved):
just a nit, but otherwise 👍

src/lib/sanity/decorators/need-lifetime-upgrade.ts
```diff
@@ -0,0 +1,49 @@
...
+      const contentPerms = new Set(item?.permission_id ?? [])
+      return contentPerms.has(MEMBERSHIP_PERMISSIONS.plus)
```
adrian-musora: This works, but it's by coincidence of how the filters are constructed. `filterBuilder` default adds base and plus content to all queries.

All plus, basic, and free, content have `MEMBERSHIP_PERMISSIONS.plus` so this query will always resolve to true. 

A better solution would be to restrict by `membership_tier === 'plus'` 

The functionality won't change, but there's a bug here that may show up later.

test/unit/lib/sanity/decorators/need-lifetime-upgrade.test.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const lifetimeOnly = { permissions: [LIFETIME], isAdmin: false, isModerator: false, isABasicMember: false }
+const lifetimePlus = { permissions: [LIFETIME, PLUS], isAdmin: false, isModerator: false, isABasicMember: false }
+const plusOnly = { permissions: [PLUS], isAdmin: false, isModerator: false, isABasicMember: false }
+const baseOnly = { permissions: [BASE], isAdmin: false, isModerator: false, isABasicMember: false }
+const admin = { permissions: [LIFETIME], isAdmin: true, isModerator: false, isABasicMember: false }
+
+const plusGatedLocked = { permission_id: [PLUS], need_access: true }
+const plusGatedUnlocked = { permission_id: [PLUS], need_access: false }
+const baseGatedLocked = { permission_id: [BASE], need_access: true }
```
adrian-musora: See my above comment about the final logic gate checking against `membership_tier` vs `permission_id`.

I don't feel like these tests were constructed with respect to how user or content permissions behave in the system. While the tests pass I think is appropriate that the tests should more closely test the situations that we'll actually see.

A user will never have just LIFETIME, all LIFETIME are associated with a BASE permission as well.
Content will never just have a BASE permission. All BASE content will also have a PLUS permission associated with it.
`baseGateLocked` is not a valid permission content  array and `'false when content does not require Plus'` is therefore nonsensical. We could add test cases that verify unexpected edge cases but not just for the sake of a test.

adrian-musora (approved):

src/lib/sanity/decorators/need-lifetime-upgrade.ts
roxanariza: Check membership_tier instead of permissions

test/unit/lib/sanity/decorators/need-lifetime-upgrade.test.ts
roxanariza: Update tests

PR #1002, opened 2026-06-30
fix(BR-671): fix Latin/World style search

Meowbox64:

- https://musora.atlassian.net/browse/BR-671

## Description/Design

- Most Styles in the search filters are 1-1 naming with their respective genre, like "Hard Rock/Metal". "Latin/World", however is stored as 2 separate genres, "Latin" and "World". This fixes queries by explicitly splitting the genre search for this case.

## Testing

- verify results show for https://devapp.musora.com:5174/guitareo/lessons/filters?filters=style,Latin/World

### DIFF
--- a/src/contentTypeConfig.js
+++ b/src/contentTypeConfig.js
@@ -877,9 +877,12 @@ function createTypeConditions(lessonTypes) {
  */
 const filterHandlers = {
   style: (value) => {
-    return (value === 'Latin/World')
-      ? (`("Latin" in genre[]->name || "World" in genre[]->name)`)
-      : (`"${value}" in genre[]->name`)
+    const hasMultipleStyles = value.includes('/')
+    const styles = [value]
+    if (hasMultipleStyles) {
+      styles.push(...value.split('/').map((s) => s.trim()))
+    }
+    return styles.map((style) => `"${style}" in genre[]->name`).join(' || ')
   },
 
   difficulty: (value) => {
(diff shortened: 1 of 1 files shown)


src/contentTypeConfig.js
```diff
@@ -876,7 +876,11 @@ function createTypeConditions(lessonTypes) {
  * Filter handler registry - maps filter keys to their handler functions
  */
 const filterHandlers = {
-  style: (value) => `"${value}" in genre[]->name`,
+  style: (value) => {
+    return (value === 'Latin/World')
+      ? (`("Latin" in genre[]->name || "World" in genre[]->name)`)
+      : (`"${value}" in genre[]->name`)
+  },
```
slumrig, line 886: just an idea but could a more general approach be used here? any style that may be delimited by '/', not sure about it being so specific to the magic string 'Latin/World'

adrian-musora: Could we fix the data instead? Is there a reason this is split into two genres but one search?

adrian-musora: I'm in favour of fixing data at the source instead of adding logical edge cases.

slumrig: I concur

Meowbox64: Chris said "Publishing had been slowly cleaning up some of the filters as there were duplicates kicking around, and I guess it got removed." so I assume this means they're working on it. I think the "any style that may be delimited by '/'" is a good solution

slumrig (approved):
looks good,

adrian-musora (approved):

PR #1009, opened 2026-07-15
fix(BEHLTP-210): add non-destructive notification polling pause

gregeatworld:

- [BEHLTP-210 — Alert notifications do not show up for manager after someone leaves plan](https://musora.atlassian.net/browse/BEHLTP-210)

## Description/Design

The web application currently stops notification polling when the document becomes hidden - for example, when the user switches to another browser tab or minimizes the browser - by calling `EventsAPI.destroy()`.

`destroy()` is a full teardown: in addition to stopping the polling interval, it removes all notification state subscribers. When polling is initialized again, unread-count requests continue, but the original subscriber no longer receives their results. As a result, the notification bell can remain out of sync even when `/notifications/v1/unread-count` returns a non-zero value.

This change introduces a non-destructive polling lifecycle:

- `pauseNotificationPolling()` stops the polling interval while retaining subscribers and initialized configuration.
- `resumeNotificationPolling()` restores the interval and immediately refreshes the unread notification state.
- Polling interval creation is idempotent to prevent duplicate intervals.
- `destroy()` remains the full teardown operation for final cleanup.

The implementation intentionally keeps the existing `initialize()`, `checkNotifications()`, subscription, and `destroy()` behavior intact.

## Testing

Automated verification:

- Added 8 notification polling lifecycle unit tests.

## Additional Notes

- This change is additive and backward-compatible.
- The default MCS export and existing public methods remain unchanged.
- Existing consumers, including mobile applications, do not need to change their current lifecycle usage.
- Consumers must explicitly opt into the new pause/resume methods.
- A follow-up frontend PR will replace temporary `destroy()`/`initialize()` calls on document visibility changes with `pauseNotificationPolling()`/`resumeNotificationPolling()`.
- `destroy()` should continue to be used for final cleanup when all polling state and subscribers must be removed.

### DIFF
--- a/src/services/eventsAPI.js
+++ b/src/services/eventsAPI.js
@@ -3,7 +3,7 @@ import {fetchLiveEvent} from "./sanity"
 import { DataContext, PollingStateVersionKey } from './dataContext'
 
 const pollingStateContext = new DataContext(PollingStateVersionKey, fetchLiveEventPollingState)
-const notificationPollingIntervalMs = 60000
+const NOTIFICATION_POLLING_INTERVAL_MS = 60_000
 
 /**
  * API for managing notifications and live event polling.
@@ -100,7 +100,7 @@ class EventsAPI {
 
     this.pollingInterval = setInterval(() => {
       this.checkNotifications();
-    }, notificationPollingIntervalMs);
+    }, NOTIFICATION_POLLING_INTERVAL_MS);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


Meowbox64 (approved):
sounds good to me

sheldonbaker (approved):

roxanariza (approved):

PR #1010, opened 2026-07-16
feat(BEHSTP-415): MCS public announcement fetch fn's

Meowbox64:

- https://musora.atlassian.net/browse/BEHSTP-415

## Description/Design

- new notifications module for this, and move notifications.js in.
- MCS side of setting up FE/MA endpoints.
- 2 endpoints:
  - get specific PA with all data like (long) `message`
  - get all PA's within period (last year) and only return data like title, date, slug

## Testing

- tests
- can also test on staging with the /devendpoint page


alesevero (approved):
Looks good to me. I'd do some testing to make sure moving the notifications file won't break FE/MA

src/services/notifications/public-announcements.ts
```diff
@@ -0,0 +1,99 @@
...
+  slug: string
+  published_on: string
+  description: string
+  message: string
+  forum_data?: Record<string, ForumData>
+}
+
+interface ForumData {
+  category_id: string
+  forum_id: string
```
adrian-musora, line 15: if you added `url` in the BE, it should also go here.
```suggestion
  forum_id: string
  url: string
```

PR #1012, opened 2026-07-20
chore(MU2-1724): add song and lesson count query

alesevero:
## Jira Ticket(s)/RelatedPR(s)

- MU2-1724

## Description/Design

- Adds `fetchSongAndLessonCounts` in `src/services/content/counts.ts` to give a single global or brand-scoped count of songs and lessons in Sanity, driven from a teammate-reviewed GROQ query.
- Uses the `Filters`/`query` builder helpers (`lib/sanity/filter.ts`, `lib/sanity/query.ts`) and `SanityClient.executeQuery` for composability and consistency with the newer infra layer, rather than the legacy `fetchSanity` GET/POST helper in `sanity.js`.
- `total` is derived in application code (`songs + lessons`) after the query returns, rather than in GROQ, to keep the query itself simple.

## Testing

- How to verify: call `fetchSongAndLessonCounts()` for a global count, and `fetchSongAndLessonCounts('drumeo')` (or another brand) for a brand-scoped count; confirm counts match manual GROQ queries run against the `production_v2` dataset.
- Environment: local
- Expected outcome: returns `{ songs, lessons, total }` with `total === songs + lessons`.

PR #1013, opened 2026-07-20
chore: make QueryBuilder stringable via toString()

alesevero:

- N/A (stacked on #1012)

## Description/Design

- Adds `toString()` to `QueryBuilder` (`src/lib/sanity/query.ts`) so a builder coerces to its built GROQ string wherever JS does implicit string coercion (template literals, `String()`, `+` concat) — same idea as PHP's `Stringable` / Java's `toString()`.
- Sweeps `artist.ts`, `genre.ts`, `instructor.ts`, and `counts.ts` to drop the now-redundant `.build()` calls at each template-literal interpolation site. Kept explicit `.build()` where a builder is passed directly as a function argument (not interpolated into a string), since that's a different call shape and the explicit call reads clearer there.
- `.build()` itself is untouched — fully additive, no breaking change to the existing API.

## Testing

- How to verify: `npx jest test/unit/lib/query.test.ts test/unit/content/counts.test.ts` — new `Stringable (toString)` describe block covers template-literal/`String()`/concat coercion and confirms it reflects live builder state, not a snapshot at creation.
- Environment: local
- Expected outcome: all tests pass; ran the full suite (70 suites) to confirm no regressions from the `.build()` removals.

## Additional Notes

- This PR targets `chore/MU2-1724-songs-lessons-counts` (#1012), not `main` — it builds on that branch's `counts.ts`. Retarget to `main` once #1012 merges.


sheldonbaker (approved):
🧶

PR #986, opened 2026-05-26
chore: logs mcs version

sheldonbaker:
log mcs version to console when imported by mpf. uses npm `prepare` hook to generate an (otherwise ignored) file containing git branch name (if `npm link`'ed) or pinned semver

PR #1015, opened 2026-07-22
fix: add gen-version script in MCS version log

sheldonbaker:
Actually adds the key version-logging that was missing in #986


slumrig (approved):
👍

PR #1017, opened 2026-07-24
chore: add subscription platform(s) to membership controller  

adrian-musora:
[MPB PR](https://github.com/railroadmedia/musora-platform-backend/pull/1406)
- Chat with Steph regarding forward solution that will support Mua data and current implementation
image

## Description/Design

- We wanted a way to understand if we should show the iframe. The current usora_users fields are based on CURRENT subscription, and these are removed in the MUA project. 
- I check the user's uaps for any apple/google product (all subscriptions by nature), and any subscription web purchases.

## Testing

Added unit tests to BE, then tested quickly with MCS cli to ensure the structure matched.


slumrig (approved):
👍

PR #1029, opened 2026-08-13
fix(TP-1323): add a condition when updating award progress

peterminhlee:
https://musora.atlassian.net/browse/TP-1323

## Testing

- Go to a GC lesson and play.
- In network tab see if awards endpoint is called multiple times

### DIFF
--- a/src/services/awards/internal/award-manager.js
+++ b/src/services/awards/internal/award-manager.js
@@ -5,6 +5,7 @@ import { AwardMessageGenerator } from './message-generator'
 import db from '../../sync/repository-proxy'
 import { STATE, COLLECTION_TYPE } from '../../sync/models/ContentProgress'
 import {getProgressStateByIds} from "../../contentProgress.js";
+import UserAwardProgressRepository from '../../sync/repositories/user-award-progress'
 
 async function getCompletionStates(contentIds, collection = null) {
   const progress = await getProgressStateByIds(contentIds, collection)
@@ -50,8 +51,8 @@ export class AwardManager {
 
   async evaluateAward(award) {
     try {
-      const hasCompleted = await db.userAwardProgress.hasCompletedAward(award._id)
-      if (hasCompleted) {
+      const existing = await db.userAwardProgress.getByAwardId(award._id)
+      if (existing?.data && UserAwardProgressRepository.isCompleted(existing.data)) {
         console.log(`Award ${award._id} already completed, skipping evaluation`)
         return
       }
@@ -63,7 +64,7 @@ export class AwardManager {
         console.log(`Award ${award._id} is now eligible, granting award`)
         await this.grantAward(award, collection)
       } else {
-        await this.updateAwardProgress(award, collection)
+        await this.updateAwardProgress(award, collection, existing)
       }
     } catch (error) {
       console.error(`Error checking award ${award._id}:`, error)
@@ -120,7 +121,7 @@ export class AwardManager {
     })
   }
 
-  async updateAwardProgress(award, collection) {
+  async updateAwardProgress(award, collection, existing) {
     try {
       const childIds = getEligibleChildIds(award)
 
@@ -145,8 +146,6 @@ export class AwardManager {
         totalLessons: childIds.length,
         completedCount
       }
-
-      const existing = await db.userAwardProgress.getByAwardId(award._id)
       const existingProgressData = existing?.data?.progress_data
       const isUnchanged = existing?.data
         && existing.data.progress_percentage === progressPercentage
@@ -154,13 +153,11 @@ export class AwardManager {
         && existingProgressData?.completedCount === progressData.completedCount
         && haveSameLessonIds(existingProgressData?.completedLessonIds, progressData.completedLessonIds)
 
-      if (isUnchanged) {
-        return
+      if (!isUnchanged) {
+        await db.userAwardProgress.recordAwardProgress(award._id, progressPercentage, {
+          progressData
+        })
       }
-
-      await db.userAwardProgress.recordAwardProgress(award._id, progressPercentage, {
-        progressData
-      })
 
       awardEvents.emitAwardProgress({
         awardId: award._id,
--- a/src/services/awards/internal/award-manager.js
+++ b/src/services/awards/internal/award-manager.js
@@ -24,6 +24,12 @@ function getCollectionFromAward(award) {
   return null
 }
 
+function haveSameLessonIds(a, b) {
+  if (!Array.isArray(a) || !Array.isArray(b) || a.length !== b.length) return false
+  const setA = new Set(a)
+  return b.every(id => setA.has(id))
+}
+
 
 export class AwardManager {
   async onContentCompleted(contentId) {
@@ -138,6 +144,18 @@ export class AwardManager {
         completedLessonIds,
         totalLessons: childIds.length,
         completedCount
+      }
+
+      const existing = await db.userAwardProgress.getByAwardId(award._id)
+      const existingProgressData = existing?.data?.progress_data
+      const isUnchanged = existing?.data
+        && existing.data.progress_percentage === progressPercentage
+        && existingProgressData?.totalLessons === progressData.totalLessons
+        && existingProgressData?.completedCount === progressData.completedCount
+        && haveSameLessonIds(existingProgressData?.completedLessonIds, progressData.completedLessonIds)
+
+      if (isUnchanged) {
+        return
       }
 
       await db.userAwardProgress.recordAwardProgress(award._id, progressPercentage, {
--- a/test/unit/awards/award-alacarte-observer.test.ts
+++ b/test/unit/awards/award-alacarte-observer.test.ts
@@ -155,7 +155,7 @@ describe('Award Observer - A La Carte Progress (null collection)', () => {
     beforeEach(() => {
       mockDb.userAwardProgress.getByAwardId.mockResolvedValue({
         data: { completed_at: new Date().toISOString(), progress_percentage: 100 }
-      })
+      } as any)
     })
 
     test('does not re-grant already completed award', async () => {
--- a/test/unit/awards/award-alacarte-observer.test.ts
+++ b/test/unit/awards/award-alacarte-observer.test.ts
@@ -153,7 +153,9 @@ describe('Award Observer - A La Carte Progress (null collection)', () => {
 
   describe('A la carte already completed award', () => {
     beforeEach(() => {
-      mockDb.userAwardProgress.hasCompletedAward.mockResolvedValue(true)
+      mockDb.userAwardProgress.getByAwardId.mockResolvedValue({
+        data: { completed_at: new Date().toISOString(), progress_percentage: 100 }
+      })
     })
 
     test('does not re-grant already completed award', async () => {
--- a/test/unit/awards/award-collection-filtering.test.ts
+++ b/test/unit/awards/award-collection-filtering.test.ts
@@ -168,7 +168,9 @@ describe('Award Collection Filtering', () => {
       expect(listeners.granted).toHaveBeenCalledTimes(2)
 
       listeners.granted.mockClear()
-      db.userAwardProgress.hasCompletedAward.mockResolvedValue(true)
+      db.userAwardProgress.getByAwardId.mockResolvedValue({
+        data: { completed_at: new Date().toISOString(), progress_percentage: 100 }
+      })
 
       emitProgress({
         contentId: 418003,
--- a/test/unit/awards/award-completion-flow.test.ts
+++ b/test/unit/awards/award-completion-flow.test.ts
@@ -197,7 +197,9 @@ describe('Award Completion Flow - E2E Scenarios', () => {
 
   describe('Scenario: User already earned the award', () => {
     beforeEach(() => {
-      db.userAwardProgress.hasCompletedAward.mockResolvedValue(true)
+      db.userAwardProgress.getByAwardId.mockResolvedValue({
+        data: { completed_at: new Date().toISOString(), progress_percentage: 100 }
+      })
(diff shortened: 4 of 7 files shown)


robwillems:
## Automated PR Review 🤖

### Context
- **Ticket:** [TP-1323](https://musora.atlassian.net/browse/TP-1323) — *Redundant awards sync pushes: `POST /api/content/v1/user/awards` fires ~2.5x more than peer routes* (Epic TP-1120 System Optimizations)
- **Modules:** `src/services/awards` (award-manager)
- **Commits:** 1
- **Agents run:** logic (JS-adapted), security, architecture, functional, test-cases
- **Skipped:** php-style, php-test-coverage — PHP-only agents, this is a JS/TS package; database — no migrations or models; claude-context — no `.claude/`, `CLAUDE.md`, `.mcp.json`, or compose changes

---

### Complexity: Small
18 lines added to a single file: a private `haveSameLessonIds` helper and a read-before-write "is this identical to what's stored?" guard in `AwardManager.updateAwardProgress`. Low structural risk, but it sits on the video-playback hot path and changes when a **public event** (`emitAwardProgress`) fires — so the blast radius is wider than the line count suggests. The new branch has effectively zero test coverage.

---

### Issues

**1. `award-manager.js:157` — The early `return` suppresses `emitAwardProgress`, not just the DB write. [High]**

The guard sits above *both* `recordAwardProgress` and `awardEvents.emitAwardProgress`. TP-1323 asks only to skip the *write*; suppressing the event is extra scope with consumers outside this repo — `emitAwardProgress` feeds `registerProgressCallback`, a public export consumed by mpf and MusoraApp (the only in-repo consumer is `test/unit/awards/award-callbacks.test.ts`).

Failure scenario: user completes 3/10 → write + emit 30%. They then open a screen whose progress ring paints only from the progress callback and does not seed itself from `getByAwardId`. Previously every playback tick re-emitted 30% and populated the ring; now nothing fires until the percentage actually changes, so the ring stays empty for the rest of the session.

Suggested fix — keep the emit unconditional:
```js
if (!isUnchanged) {
  await db.userAwardProgress.recordAwardProgress(award._id, progressPercentage, { progressData })
}
awardEvents.emitAwardProgress({ awardId: award._id, progressPercentage, timestamp: Date.now() })
```
If suppressing the event *is* intentional, please say so in the PR body — the stated test plan (watch the network tab) would not catch a UI regression from it.

**2. `award-manager.js:149` — The guard's effectiveness depends on the server echoing `progress_data` back unchanged. If it doesn't, this ships as a silent no-op. [High — verify before merge]**

After a successful push, `store.executePush` → `writeEntries(successfulEntries)` → `prepareRecords` overwrites the local raw columns with the server's response (`r._raw[key] = value`). Same on any pull. So the record the guard reads on tick N+1 is the *server's* version of `progress_data`, not the one the client wrote.

Failure scenario: the backend doesn't persist `progress_data`, or normalises it to snake_case (`total_lessons` / `completed_count`). The echo overwrites local raw; on the next tick `existingProgressData?.totalLessons` is `undefined` while `progressData.totalLessons` is a number → `isUnchanged` is false → every tick still writes, and the endpoint is called exactly as often as before. Nothing in CI would catch this.

Mitigating: the comparison is robust to JSON re-encoding and to `completedLessonIds` reordering, so a verbatim-ish echo is fine. Please confirm against the backend that `progress_data` round-trips with its camelCase keys intact.

**3. `award-manager.js:149` — Redundant second read of the same record in the same call stack. [Medium]**

`evaluateAward` already fetched this record via `hasCompletedAward(award._id)` → `readOne(award._id)`. The new `getByAwardId` repeats that work verbatim: a `db.localStorage.get` fetch-token check, a WatermelonDB query, and a full `ModelSerializer.toPlainObject` (which invokes every getter, including `JSON.parse` of `progress_data` and `completion_data`).

On React Native, every debounced progress tick for a learning-path award now does two AsyncStorage round-trips and two full record serialisations instead of one, for zero new information. Read once in `evaluateAward`, derive `isCompleted` from it via `UserAwardProgressRepository.isCompleted`, and pass the record into `updateAwardProgress`.

(Not a network concern — `_respondToRead` only forces a `pullRecords()` when the fetch token is empty, and `hasCompletedAward` always runs first, so any first-time pull has already happened.)

**4. Zero behavioural coverage — every existing awards test exercises only the pre-PR path. [Medium]**

`test/unit/awards/helpers/mock-setup.ts:18` declares `getByAwardId: jest.fn()` and never gives it a return value, and `setupAwardMocks` doesn't configure it. So it resolves `undefined` → `existing?.data` is falsy → `isUnchanged` is falsy in **every** test across `award-multi-lesson`, `award-completion-flow`, `duplicate-prevention`, `award-collection-edge-cases`, and `award-observer-integration`. All of them pass whether the guard is correct, broken, or inverted. Combined with Codecov's 30.76% patch coverage, a regression here is undetectable. The missing assertion is two lines — see Test Cases below.

---

### Test Coverage
Codecov: **30.76% patch coverage — 8 lines missing, 1 partial.** No tests added.

The gap is not incidental: as above, the shared awards mock leaves `getByAwardId` returning `undefined`, so the entire new branch plus `haveSameLessonIds` is dead code under the existing suite. Suggested home: `test/unit/awards/award-progress-dedupe.test.ts`, reusing `test/unit/awards/helpers/mock-setup.ts`. `haveSameLessonIds` is module-private — export it, or test it through `updateAwardProgress`.

If you give `setupDefaultMocks` a default of `{ data: null }` for `getByAwardId`, re-run the full awards suite — that default is what keeps the new tests from passing for the wrong reason.

---

### Test Cases

| # | Test Case | Steps | Expected Result | Type |
|---|-----------|-------|-----------------|------|
| 1 | Unit: repeated identical ticks write once | Mock 1 of 3 lessons complete (33%); stub `getByAwardId` → `{ data: { progress_percentage: 33, progress_data: { completedLessonIds: ['l1'], totalLessons: 3, completedCount: 1 } } }`; call `updateAwardProgress` twice | `recordAwardProgress` called 0 times; no sync push queued | Happy path |
| 2 | Unit: a genuinely new completion still writes | Same stored record; now `l1` + `l2` complete (67%) | `recordAwardProgress(award._id, 67, { progressData: {...completedCount: 2} })` called once; progress event emitted at 67% | Happy path |
| 3 | Unit: first-ever write when no stored record | `getByAwardId` → `undefined`, then `{ data: null }`; 1 of 3 complete | Both write once at 33% — this is the path every existing test hits today, must keep passing | Regression |
| 4 | Unit: `haveSameLessonIds` rejects non-arrays / length mismatch | `(undefined, ['a'])`, `(null, [])`, `('a,b', ['a','b'])`, `(['a'], ['a','b'])`, `([], [])` | First four `false`, `([], [])` `true`; via `updateAwardProgress` a missing/non-array stored value always fails open to a write | Edge case |
| 5 | Unit: order-insensitive comparison | `(['l3','l1','l2'], ['l1','l2','l3'])`; and store `['l3','l1']` vs computed `['l1','l3']` at equal percentage | `true`; `recordAwardProgress` not called — reordering alone must not re-push | Edge case |
| 6 | Unit: stored record with null/partial `progress_data` still writes | `{ progress_percentage: 33, progress_data: null }`, then `{ progress_data: { completedLessonIds: ['l1'] } }` (no counts) | Both fall through the guard and write once, repairing the incomplete record | Edge case |
| 7 | Manual QA: GC playback fires awards endpoint only on real change | Open a GC lesson in an award, DevTools → Network filtered to the awards endpoint; play 60s+ without completing; then complete it and play/seek another 60s; then reload and replay the partially-watched lesson | ≤1 write during the first 60s (not one per tick); exactly one more after the completion; **zero** after reload; progress UI still correct throughout | Performance |
| 8 | Manual QA: completing the final lesson still grants the award | On a path with one lesson left, complete it; then replay/re-complete it | 100% write with `completedAt` + `completionData`, granted popup, certificate shows earned; re-completing does not re-grant (short-circuited by `hasCompletedAward`) | Regression |

---

### Functional
**Verdict: Partially aligned.**

TP-1323's core directive — *"skip the write when the computed progress equals the stored value"* — is implemented, on the exact hot path the ticket identified. It should remove the bulk of the redundant POSTs, and it correctly handles the "one lesson belongs to multiple awards" multiplier since each award is compared independently.

Gaps against the ticket:
- ✓ Skip the write when computed progress equals stored value
- ✓ Handles the multi-award multiplier
- ✗ **The ticket's explicit hint — add a `comparator` for `UserAwardProgress` in `store-configs.ts` — is not addressed.** `UserAwardProgress` is still the only store besides `ContentLike` without one. (The hint is arguably a partial misdiagnosis — `ContentProgress`'s comparator returns `'SERVER' | 'LOCAL'`, i.e. pull-side conflict resolution, not push dedup — so the source-level guard is a defensible substitute for the primary requirement. But please confirm the comparator half is deliberately deferred rather than overlooked: server pulls re-dirtying local award records is a second path to the same symptom, still untouched.)
- ✗ **Fix is at one call site, not the write layer the ticket named** (`user-award-progress.ts:106`, `upsertOne` writes unconditionally). `grantAward` and `completeAward` remain unconditional — both are gated upstream by `hasCompletedAward` so real-world redundancy there is low, but this is now a per-call-site convention the next caller can silently skip.

---

### Architecture

**The guard belongs one layer down, in the sync core.** `SyncStore.upsertSome` (`src/services/sync/store/index.ts`) unconditionally calls `existing.prepareUpdate(builder)`, marking the record `_status: 'updated'` and bumping `updated_at` even when no field value actually changed. It then fires `pushUnsyncedWithRetry`, and `SyncRepository._respondToWrite` additionally fires `pushRecordIdsImpatiently` — a real, un-throttled HTTP push. A semantically no-op write becomes a network call.

A "no raw field changed → skip the batch and skip the push" check inside `upsertSome` would fix this once for `content-progress.ts`, `content-likes.ts`, `practices.ts`, and awards. Notably `content-progress.ts`'s `upsertOne` sits on the *same* video-playback hot path with the identical problem, untouched here.

Other notes:
- **`haveSameLessonIds` naming/placement** — a generic order-insensitive set-equality with a domain-specific name, private to a domain file. If the guard stays here it belongs in `src/services/sync/utils/` or `src/lib/` under a generic name; if it moves to the sync layer it disappears in favour of a generic raw-field comparator, which is the better outcome.
- **Offline-first behaviour change** — repeated no-op writes previously acted as an incidental re-push trigger for a record left dirty after a failed push (the `SyncError` from `_respondToWrite` is swallowed by `updateAwardProgress`'s `catch`). Such a record now relies solely on the sync strategies (hourly Polling / Online / Activity). The guard compares **local** state only — local holding the value is not evidence the server has it. Defensible, but it's a change to the retry surface made from above the layer that owns retry.
- **Debounce is not an alternative** — `content-progress-observer.js`'s `debounceMs = 50` is far below the cadence of `onProgressSaved` during playback, so it coalesces essentially nothing. But even a correctly-tuned debounce wouldn't stop identical writes, because the underlying problem is that unchanged data still produces a push. Complementary, not a substitute.

---

### Security
✓ No security issues.

No auth, permission, or `need_access` logic changed. No new network calls, headers, endpoints, or credential handling — the only added I/O is a local WatermelonDB read of the user's own award-progress row, which this same class already reads via `hasCompletedAward`. `haveSameLessonIds` does `Array.isArray` checks, a `Set` build, and `every` — no dynamic key assignment (no prototype-pollution vector), no regex, no DOM sink. Nothing new logged.

---

### Notes

- **`haveSameLessonIds` is multiset-blind but not exploitable here.** `new Set(a)` + `b.every(...)` treats `['l1','l1','l2']` and `['l1','l2','l2']` as equal. Reaching that requires duplicate ids in `award.child_ids` (Sanity's `content->child[...]->railcontent_id` doesn't dedupe), but duplicates always flip completion together, so the completed multiset can't differ while `completedCount` — compared on the line above — stays equal. Not a defect, just worth knowing the semantics are *set* equality, not array equality. Minor: the `a.length !== b.length` check is dead weight, since `completedCount` is already compared and equals `completedLessonIds.length` by construction.
- **Falsy/edge handling is correct.** `readOne` returns `data: null` when absent (and `findRecord` excludes soft-deleted records, so a pending `deleteAllAwards` can't resurrect stale progress). `progress_percentage === 0` is safe — strict comparison, not truthiness. First-ever write is safe. A record with `progress_data: null` (e.g. written by `completeAward`) is safe: `null?.totalLessons` is `undefined` and `progressData.totalLessons` is always ≥ 1 because `childIds.length === 0` returns early.
- **No concurrency gap.** `processingContentIds` in `ContentProgressObserver` serialises `checkAward` per award content id, so the read-check-write across `await` boundaries can't interleave for the same record.
- **Pre-existing, not introduced here:** the per-tick work still runs in full even when the write is skipped — `getProgressStateByIds` is called twice per evaluation (once in `checkAwardEligibility`, again in `updateAwardProgress`). And `pushUnsyncedWithRetry` sweeps *all* unsynced rows in the table on any write, so one genuine change still flushes everything.

---
*Automated review — see `.claude/skills/pr-review/SKILL.md`*


[TP-1323]: https://musora.atlassian.net/browse/TP-1323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

slumrig (commented):
Nice fix, this should cut down on the redundant awards calls nicely!

One thing: none of the updated tests actually exercise the new isUnchanged skip path. They all cover the "already completed" case or the default "needs update" case, but not "progress recomputed the same as existing, so recordAwardProgress should be skipped." Mind adding a test that seeds getByAwardId with progress data matching what gets recomputed, then asserts recordAwardProgress wasn't called? This would ensure this doesn't quietly regress later.

calebfavor (approved):

PR #1030, opened 2026-08-13
feat(BEHLTP-306): add tier_key and rank to UpgradeProduct type

slumrig:

- https://musora.atlassian.net/browse/BEHLTP-306
- Related: musora-platform-backend PR #1463

## Description/Design

- The backend's `upgrade_options` response now includes `tier_key` and `rank` per product (added in BEHLTP-306 / PR #1463), so MusoraApp can fetch canonical plan rank from the API instead of hardcoding its own copy.
- This adds the matching `tier_key: string` and `rank: number | null` fields to MCS's `UpgradeProduct` TypeScript interface, so consumers (e.g. MusoraApp) can reference these fields without a type error.

## Testing

- How to verify: `npx tsc --noEmit` passes cleanly with the new fields added.
- Environment: local
- Expected outcome: no type errors; `product.tier_key`/`product.rank` are now valid, typed accesses wherever `UpgradeProduct` is consumed.

### DIFF
--- a/src/services/user/memberships.ts
+++ b/src/services/user/memberships.ts
@@ -75,6 +75,8 @@ export interface UpgradeProduct {
   tier: 'plus' | 'basic' | '' // deprecated in favour of membership_level
   membership_level: 'plus' | 'basic' | ''
   plan_type: 'family' | 'duo' | 'solo'
+  tier_key: string
+  rank: number | null
 }
 
 export interface UpgradeOption {
--- a/.agent/decisions/2026-08-13-behltp-306-add-plan-rank-and-tier-key.md
+++ b/.agent/decisions/2026-08-13-behltp-306-add-plan-rank-and-tier-key.md
@@ -1,0 +1,24 @@
+---
+date: 2026-08-13
+branch: feat/BEHLTP-306-add-plan-rank-and-tier-key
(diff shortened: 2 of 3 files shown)


src/services/user/memberships.ts
```diff
@@ -75,6 +75,8 @@ export interface UpgradeProduct {
   tier: 'plus' | 'basic' | '' // deprecated in favour of membership_level
   membership_level: 'plus' | 'basic' | ''
   plan_type: 'family' | 'duo' | 'solo'
+  tier_key: string
+  rank: number | null
```
Meowbox64, line 79: as a more general note, should we test interfaces to ensure certain fields are sent, and that the interface doesn't regress?

Meowbox64 (approved):

src/services/user/memberships.ts
slumrig: Good catch. This interface has zero runtime enforcement right now. If the backend drops or renames a field, or if someone edits the interface incorrectly, nothing catches it. MCS already has a test:live tier (jest.live.config.js) hitting real endpoints, so a live test asserting the actual upgrade_options response includes tier_key/rank would catch drift in either direction. It's a broader testing strategy question with multiple systems.

PR #1031, opened 2026-08-17
fix: password reset mobile token and sessionConfig access

alesevero:

- N/A

## Description/Design

- `resetPassword` now accepts optional `deviceName`, `deviceToken`, `platform` so mobile can exchange a reset token for an auth session in one call; also fixed a bug sending `password` as the `token` field.
- `listOAuthProviders`/`unlinkOAuthProvider` read `userId` from `globalConfig.sessionConfig` instead of a stale top-level `globalConfig.userId`, matching current config shape.

## Testing

- How to verify: call `resetPassword` with device fields set and confirm the mobile app receives an `AuthResponse` session back; call `listOAuthProviders`/`unlinkOAuthProvider` and confirm userId resolves correctly.
- Environment: local
- Expected outcome: password reset returns a valid session on mobile; OAuth provider list/unlink calls succeed instead of hitting `undefined` userId.

### DIFF
--- a/.agent/decisions/2026-08-17-password-reset-mobile-token.md
+++ b/.agent/decisions/2026-08-17-password-reset-mobile-token.md
@@ -1,0 +1,27 @@
+---
+date: 2026-08-17
+branch: fix/password-reset-mobile-token
+pr: https://github.com/railroadmedia/musora-content-services/pull/1031
+status: open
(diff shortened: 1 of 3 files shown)


slumrig (approved):
Image

PR #1035, opened 2026-08-19
fix: escape values interpolated into GROQ filter strings

alesevero:

- N/A

## Description/Design

`Filters` built its expressions by interpolating caller-supplied values straight into GROQ string literals, so a value containing a double quote closed the literal early and had the remainder parsed as GROQ.

**The reachable path is search.** `fetchInstructorLessons`, `fetchArtistLessons` and `fetchGenreLessons` are all exported and all pass their `searchTerm` option into `f.searchMatch('title', searchTerm)` (`instructor.ts:144`, `artist.ts:144`, `genre.ts:143`). `searchTerm` comes from the search box, so:

```ts
Filters.searchMatch('title', 'x" || _type == "user')
// before → title match "x" || _type == "user*"
```

The `||` is now GROQ, not text — the filter matches documents of a type the caller never scoped to.

### The fix

Two helpers, applied to every value interpolated into a string literal:

```ts
const escapeDoubleQuoted = (value: string | number): string =>
  String(value).replace(/\\/g, '\\\\').replace(/"/g, '\\"')

const escapeSingleQuoted = (value: string | number): string =>
  String(value).replace(/\\/g, '\\\\').replace(/'/g, "\\'")
```

Backslash is escaped first so a trailing `\` cannot escape the closing quote.

**All 11 interpolation sites are fixed, not just the two search ones.** `searchMatch`/`titleMatch` were the reachable path, but `brand`, `type`, `slug`, `references`, `referencesField`, `publishedBefore`, `publishedAfter` and the single-quoted items inside `typeIn`/`statusIn` had the identical defect. Patching only the reported path would have left every sibling caller broken.

`railcontentId` and `idIn` are coerced with `Number` instead of escaped. They're typed `number`, but MCS is consumed from JavaScript too, so a string argument would have injected an expression directly (`railcontent_id == 1 || true`). They now yield `NaN` and match nothing.

### Not changed

Field-name parameters (`searchMatch(field, …)`, `referencesField(field, …)`, `defined(field)`) are GROQ **identifiers**, not string literals — they can't be escaped this way, and they're supplied by code in this repo rather than by users.

## Testing

- How to verify: `npm test -- test/unit/lib/filter.test.ts`
- Environment: local
- Expected outcome: 129/129 pass. Full suite 1183 pass, `tsc --noEmit` and `prettier --check` clean.

8 new cases, asserting the actual payloads:

```ts
test('a quote in a search term cannot close the string literal', () => {
  const injection = 'x" || _type == "user'
  expect(Filters.searchMatch('title', injection)).toBe('title match "x\\" || _type == \\"user*"')
})

test('a trailing backslash cannot escape the closing quote', () => {
  expect(Filters.titleMatch('rock\\')).toBe('title match "rock\\\\*"')
})
```

## Additional Notes

- **Behaviour-preserving for legitimate input.** Escaping is a no-op on values without quotes or backslashes, so **no pre-existing assertion changed** across the 129 filter tests — including the unicode cases (`Señor`, `日本語`). A test pins that explicitly.
- Worth a look at whether any stored content slugs or titles contain a literal `"` or `\`; those queries were previously malformed and will now be correctly quoted, which changes their results from "broken" to "right".

### DIFF
--- a/src/lib/sanity/filter.ts
+++ b/src/lib/sanity/filter.ts
@@ -66,8 +66,14 @@ const applyPrefix = (prefix: Prefix, filter: string): string => {
   return filter.replace(/\b([a-z_][a-z0-9_]*)\s*(==|!=|<=|>=|<|>|in|match)/gi, `${prefix}$1 $2`)
 }
 
+const escapeDoubleQuoted = (value: string | number): string =>
+  String(value).replace(/\\/g, '\\\\').replace(/"/g, '\\"')
+
+const escapeSingleQuoted = (value: string | number): string =>
+  String(value).replace(/\\/g, '\\\\').replace(/'/g, "\\'")
+
 const arrayToStringRepresentation = (arr: string[]): string => {
-  return '[' + arr.map((item) => `'${item}'`).join(',') + ']'
+  return '[' + arr.map((item) => `'${escapeSingleQuoted(item)}'`).join(',') + ']'
 }
 
 // ============================================
@@ -87,7 +93,7 @@ export class Filters {
    * @example Filters.brand('drumeo') // "brand == 'drumeo'"
    */
   static brand(brand: string): string {
-    return `brand == "${brand}"`
+    return `brand == "${escapeDoubleQuoted(brand)}"`
   }
 
   /**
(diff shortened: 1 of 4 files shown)


slumrig (approved):
Image

PR #1036, opened 2026-08-19
feat: make built GROQ queries runnable with a typed result

alesevero:

- Builds on #1034 (`composite()`), already merged
- Plan: `.claude/docs/architecture/groq-query-builder-execution.md` (tracked by this PR)

## Description/Design

The query builder built GROQ strings and stopped there — `build()` returned a string and nothing ran it. Execution lived in `SanityClient`, connected only by string interpolation.

The error contract was the thing worth fixing. All three `SanityClient` methods *look* like they return a null-ish fallback on failure:

```ts
} catch (error: any) {
  this.handleError(error, query)
  return null            // SanityClient.ts:37, :52 (`return []`), :67
}
```

`handleError` is declared `: never` and throws unconditionally, so **all three fallback returns are unreachable**. The real contract is "`null` when the query matched nothing, throws when the request failed" — two very different outcomes, one invisible at the type level.

### What this adds

```ts
const result = await groq()
  .and(f.type('song'))
  .order('published_on desc')
  .slice(0, 10)
  .run<Song[]>()

result.fold(
  (error) => console.error(`${error.name}: ${error.message}`, error.groq),
  (songs) => songs ?? []
)
```

- **`run()` returns `Either<SanityQueryError, T | null>`** instead of throwing. A query that matched nothing is `Right(null)`; a failed request is `Left`. That distinction is the entire point, so empty is deliberately *not* folded into `Left`.
- **`SanityQueryError` is a real `Error` subclass**, replacing the object literals cast to an interface that the Sanity infrastructure throws today — those carry no stack and fail `instanceof Error`.
- **`groq()` is a composition root**, in its own module. It is *not* a method on `QueryBuilder`: `filter.ts` imports `query.ts`, so a `run()` method there would drag the transport into every consumer of `Filters`. `groq.ts` imports both `query` and `runner`; neither imports the other.
- **`groq.composite()`** wraps the merged `composite()` into a `RunnableQuery` that is deliberately **not** chainable — `.and()` on a composite isn't valid GROQ, and the type now says so.
- **`mapAsync` on `Coproduct`**, so the existing typed decorators compose with query results (see below).

### Decorators compose through the ADT

The decorators in `src/lib/sanity/decorators/` are `(items) => items` transforms, so the sync ones already compose with `map`. `decorateNavigateTo` is async, and `map` with an async function yields `Either<E, Promise<T>>` — a promise trapped inside the container. Nothing in the repo offered `mapAsync`/`traverse`/`sequence`, so a caller had to abandon the ADT exactly where it should help.

```ts
const decorated = await result
  .map((lessons) => decorateAccess(lessons ?? [], permissions))   // WithNeedAccess<Lesson>[]
  .mapAsync((lessons) => decorateNavigateTo(lessons))             // + WithNavigateTo
```

A rejection propagates rather than becoming a `Left` — there is no way to build an `L` from it, and the never-throws guarantee belongs to the boundary, not to the functor. A test asserts the propagation so it isn't later mistaken for an oversight and swallowed.

**No decorator file is modified.** `git diff --name-only main...HEAD -- src/lib/sanity/decorators/ src/services/` is empty for this PR.

### Deliberate omissions

- **No `kind`/`status` on the error.** `createSanityError` (`FetchQueryExecutor.ts:89-109`) formats the HTTP status into a message string and returns an object without it — and overwrites that string entirely when Sanity returns a JSON body. Neither is obtainable without editing the transport, which this PR doesn't touch.
- **No `params` argument.** `Filters` bakes values in before `run()` is called, so nothing could supply one.
- **No `runComposite`.** It would be a pure alias for `run(composite(parts))`.
- **No decorate-and-run helper.** One was written and removed: it serves no call site yet, and shaping it without one produced an API whose options were guesses. The two things it encoded — fetching permissions alongside the query, and catching a rejected permissions fetch into an `Either` before `Promise.all` sees it — are demonstrated in `examples.ts` instead. Worth revisiting when a real migration needs it.

### `Either` is an alias, not a subclass

```ts
export type Either<L, R> = Coproduct<L, R>
export const Either = { left: Coproduct.left, right: Coproduct.right }
```

Categorically the binary coproduct **is** `Either` — `Left`/`Right` are its canonical injections, which is why Haskell and Scala declare them on `Either` directly. Subclassing would carve at the wrong joint: neighbouring types differ by *algebra*, not shape (`Validation` accumulates errors applicatively; `These` has three cases). Shared behaviour belongs on the existing typeclasses — `Functor`, `Monad`, `Tappable`, `Foldable`, `Recoverable`, `Droppable`.

`Coproduct` itself is ported forward from the long-lived `chore/ads-structures` branch, which sits 1331 commits behind main and has no `src/lib/sanity`, so nothing on it was usable as-is.

## Testing

- How to verify: `npm test`
- Environment: local
- Expected outcome: 1227 pass across 75 suites. `tsc --noEmit` and `prettier --check` clean.

The tests carrying design rather than code:

```ts
// runner.test.ts — the lazy singleton must not pin stale config
test('reads the configuration again after globalConfig changes', ...)

// groq.test.ts — opposite build semantics, pinned as a matched pair
test('runs the query as built at call time', ...)                 // groq()
test('stringifies its parts eagerly, unlike a groq() chain', ...) // groq.composite()

// coproduct.test.ts — the documented contract, so nobody "fixes" it into a swallow
test('propagates a rejection instead of converting it to a left', ...)
```

Fakes construct a **real** `SanityClient` with fake `QueryExecutor`/`ConfigProvider` rather than mocking `executeQuery`, so the actual `handleError` conversion is exercised — that conversion is the thing being built.

## Additional Notes

- **Ships unused.** No service file is modified; `query.ts`, `filter.ts`, `SanityClient.ts`, `FetchQueryExecutor.ts`, everything under `decorators/`, and `services/sanity.js` are untouched. Call sites migrate one at a time in later PRs.
- **`examples.ts` earned its place twice.** It's compile-checked usage (precedent: `infrastructure/sanity/examples/usage.ts`, `lib/sanity/decorators/examples.ts`). Writing it surfaced (1) that `QueryRunner` put its generic on the *call* rather than the type, making the runner impossible to fake without `as any` — which three unit tests had already silently done; and (2) that `WithNeedAccess<T>` degrades the recursive `children` type when `T` has an index signature, breaking a `decorateAccess` → `decorateNavigateTo` chain. JSDoc `@example` blocks would have caught neither; they aren't compiled.
- Example files are now excluded from coverage collection — no business logic, not exported from the package index, so counting them as uncovered production code was noise. This *raises* the reported percentage rather than lowering the thresholds.
- **Config freshness.** The default runner is a lazy singleton calling `refreshConfig()` per query, because `initializeService` mutates `globalConfig` in place while `DefaultConfigProvider` hands back a copy that `SanityClient` caches — otherwise it serves a snapshot taken at first query, which breaks between test cases.
- **Known migration blockers**, recorded for whoever moves a `fetchSanity` call site: results from `run()` are undecorated by default; `decorateAll` only walks `children`, whereas the legacy `contentResultsDecorator` also walks `result.content[]` and `results.entity[].lessons`; and `fetchUserPermissions` memoizes *rejected* promises for up to 10s, so one blip can produce a run of failures.
- Includes the Coproduct port that was on `feat/ads-coproduct`; that branch gets no separate PR.

### DIFF
--- a/.claude/docs/architecture/groq-composite-projections.md
+++ b/.claude/docs/architecture/groq-composite-projections.md
@@ -1,0 +1,223 @@
+# Plan: `composite()` — multi-key projections in the query builder (mcs)
+
+Status: proposed, not implemented.
+Scope: `src/lib/sanity/query.ts`.
+Related: `.claude/docs/architecture/groq-query-builder-execution.md` — independent of it; see "Relationship to the
+execution plan" below.
+
+## Context
+
+`src/lib/sanity/query.ts` has one projection slot. It can emit `*[...]{ a, b }` but it cannot
+emit a wrapper object whose values are themselves queries:
+
+```groq
+{ "data": *[...]{...}, "total": count(*[...]) }
+```
+
+That wrapper is the dominant idiom in `src/services/content/`. Because the builder can't express
+it, every such call site drops out of the builder and hand-writes a template literal, then calls
+the transport directly. Seven sites do this today.
+
+The cost is that the builder is bypassed exactly where queries are most complex, and its output
+gets spliced into strings by hand — including one site that relies on implicit `toString()`
+(`counts.ts:37`, `${query().and(songsFilter)}` with no `.build()`) while a sibling calls
+`.build()` explicitly (`artist.ts:88`). Same idiom, two spellings, no enforcement.
+
+Goal: let the builder express the wrapper, so those sites can use it.
+
+## What the call sites actually do
+
+Across all seven builder-based composite sites there are exactly **two** value forms:
+
+| form | sites |
+|---|---|
+| `"key": <builder>` | `artist.ts:55`, `genre.ts:55`, `instructor.ts:56`, plus the `"data"` half of the three below |
+| `"key": count(<builder>)` | `artist.ts:158-159`, `genre.ts:157-158`, `instructor.ts:157-158`, `counts.ts:37-38`, `counts.ts:62` |
+
+```ts
+// counts.ts:36-39 — the only builder → SanityClient path in the repo
+const q = `{
+  "songs": count(${query().and(songsFilter)}),
+  "lessons": count(${query().and(lessonsFilter)})
+}`
+
+// instructor.ts:156-159 (artist.ts:157-160 and genre.ts:156-159 identical in shape)
+const q = `{
+  "data": ${data},
+  "total": count(${total})
+}`
+```
+
+Two forms, one level deep, string keys. That is the entire requirement.
+
+## The `count(...)` half needs no new code
+
+`Filters.count` already exists (`src/lib/sanity/filter.ts:460`):
+
+```ts
+static count(filter: string): string {
+  return `count(*[${filter}])`
+}
+```
+
+It is exactly equivalent to what the call sites hand-roll. `build()` ends in `.trim()`
+(`query.ts:149`), and for a bare `query().and(x)` the projection, post-filter, ordering and slice
+slots are all empty — they collapse to whitespace-only trailing lines that `trim()` removes:
+
+```
+count(${query().and(songsFilter)})   ===   count(*[songsFilter])   ===   f.count(songsFilter)
+```
+
+So `composite()` must **not** grow a `count()` wrapper of its own. Count values arrive as plain
+strings from the existing helper. That collapses the new surface to the one thing genuinely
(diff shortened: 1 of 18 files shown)


slumrig (approved):
Think I got a vague jist of this. Looks good!

